PR #2, opened 2017-04-04
Update webpack config

jmcolella:


jmcolella:
@drewdrewthis 👀  please. webpack config stuff

PR #7, opened 2017-04-06
Add beneficiary and textlockup

drewdrewthis:


src/components/text-lockups/text-lockups.module.scss
```diff
@@ -0,0 +1,22 @@
+@import 'src/assets/stylesheets/base_styles/typography';
+@import 'src/assets/stylesheets/base_styles/colors';
+@import 'src/assets/stylesheets/utilities/functions';
+@import 'src/assets/stylesheets/utilities/mixins';
+
+.type1 {
+  $subheader-color: $neutral-4;
+
+  .header {
```
jmcolella: what's the benefit of nesting here? I don't think we need to do this in general, just maybe for pseudo selectors

src/modules/cards/cards.module.scss
```diff
@@ -51,18 +52,98 @@ $hover-color:        $neutral-2;
...
+  position: relative;
+  min-width: ru(11.33);
+  background-color: $beneficiary-card-background-color;
+  margin-bottom: ru(1);
+
+  .mod-percentage-card {
+    display: none;
+  }
+
+  &.collapsed {
```
jmcolella: this should not be nested but instead composed from beneficiary card. `composes` will do the work of `&` behind the scenes.

This opens up the discussion of making individual Sass files for each module/component

src/modules/cards/cards.module.scss
```diff
@@ -1,24 +1,25 @@
-@import 'src/assets/stylesheets/base_styles/typography';
-@import 'src/assets/stylesheets/base_styles/colors';
-@import 'src/assets/stylesheets/utilities/functions';
-@import 'src/assets/stylesheets/utilities/mixins';
+@import 'src/assets/stylesheets/global';
```
jmcolella: smart 👍

drewdrewthis: @jmcolella Kind of.. but do we want to do it explicitly, or do we want it to simply be part of our post-css pipeline?

src/modules/cards/cards.module.scss
drewdrewthis: @jmcolella Agreed -- these styles were actually pulled in from our existing repo

src/modules/cards/cards.module.scss
jmcolella: @drewdrewthis yeah I had that thought in the back of my mind as well...I'm going to be investigating more into post css today, so we should have a better idea

PR #8, opened 2017-04-07
Add List & ArticleList

cisacke:


src/assets/stylesheets/utilities/functions.scss
```diff
@@ -32,3 +32,17 @@ $rem-ratio:  24px; // Line height
...
+
+@function lower-bound($range){
+  @if length($range) <= 0 {
+    @return 0;
+  }
+  @return nth($range,1);
+}
+
+@function upper-bound($range) {
+  @if length($range) < 2 {
```
drewdrewthis: Gosh, these foundation mixins are so complicated.

src/components/lists/ArticleList/index.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <li>
+      <a href={ link }>
+        <figure>
+          <img src={ imageSrc } />
+
+          <figcaption>
+            <h3 className={styles.header}>{ header }</h3>
+            <p>{ subHeader } </p>
```
drewdrewthis: Is this space after `{ subHeader}` intentional?

src/components/lists/ArticleList/index.js
```diff
@@ -0,0 +1,35 @@
...
+  } = props;
+
+  return (
+    <li>
+      <a href={ link }>
+        <figure>
+          <img src={ imageSrc } />
+
+          <figcaption>
+            <h3 className={styles.header}>{ header }</h3>
```
drewdrewthis: We should standardize our paren/bracket spacings.

src/components/lists/List/index.js
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
-import styles from './lists.module.scss';
+import styles from '../lists.module.scss';
 
 function List ( { type, children } ) {
   return (
-    <div className={ styles[type] }>{children}</div>
+    <ul className={ styles[type] }>
+      {children}
```
drewdrewthis: This would require a dev to then add `<li>` as children. I'm thinking that we should make this easier for them and have this maybe be the default unless they pass in an array of text as a prop, in which case the component would build itself with that given array.

src/components/lists/lists.module.scss
```diff
@@ -0,0 +1,78 @@
...
+li {
+  margin-bottom: ru(1);
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+}
+
+.header {
+  @include like-h5;
```
drewdrewthis: I'm ripping this `like-h5` mixin out in my PR. IMO we shouldn't be using it, we should just be using the typography mixins.

src/components/lists/lists.module.scss
```diff
@@ -0,0 +1,78 @@
...
+  margin: 0;
+  color: $article-list-heading-color;
+}
+
+p {
+  @include like-p;
```
drewdrewthis: Same as above -- also you shouldn't be setting the font-family here -- this is something that should be set via the typography mixin.

Also, I think there's a strong enough argument against setting styles on HTML elements that we should follow it and use classes only.

src/components/lists/lists.module.scss
```diff
@@ -0,0 +1,78 @@
...
+a {
+  text-decoration: none;
+}
+
+figure {
+  margin: 0;
+  padding: 0;
+}
+
+img:not([width]) {
```
drewdrewthis: What is this?

cisacke:
@drewdrewthis made those changes - ready to merge if this looks good to you

src/components/lists/ArticleList/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const defaultItemProps = [ 'id', 'link', 'imageSrc', 'header', 'subHeader' ];
+  const itemProps = reduce(defaultItemProps, (res, val) => {
+    res[val] = mapProps[val] || val;
+
+    return res;
+  }, {});
+
+  return (
+    <List type={ type } >
```
drewdrewthis: @cisacke Why did you abstract `List` into a component instead of just using a `ul` tag?

cisacke: Hmm yeah it was like this originally because we were planning to expose it as a wrapper component to the end user, but in the current iteration I agree it can just be a `ul` - I'll make the change

PR #12, opened 2017-04-10
Add formfields and add weight mixins to typography

drewdrewthis:


src/assets/stylesheets/global.scss
```diff
@@ -1,8 +1,8 @@
+@import 'normalize';
```
jmcolella: so right now, this normalize (and any 'global' selectors) is a major issue. every time we import global into another scss file, it brings in ALL of these classes and assigns them module names

this means, if we have 5 files importing `global`, then all of the `global` selectors will be included in the final css 5 times, causing huge bloat

we can discuss as a team solution

jmcolella:
@drewdrewthis LGTM other than my note on imports; that's a bigger issue but doesn't block this PR as is

src/components/formfields/FormField.js
```diff
@@ -0,0 +1,120 @@
...
+
+FormField.propTypes = {
+  /**
+   * Possible card types are: `1`
+   *
+   */
+  type: React.PropTypes.string.isRequired,
+  forProp: React.PropTypes.string,
+  /**
+   * Label is optional. If not provided, component will reorganize accordingly.
```
cisacke: Is the expected behavior for no label (and default `type`) a Simple Input? If so I think you'd need to remove the `label` tag entirely

src/components/formfields/FormField.js
```diff
@@ -0,0 +1,120 @@
...
+            type='text'
+            placeholder={ placeholder } />
+        </div>
+      )
+  }
+}
+
+FormField.propTypes = {
+  /**
+   * Possible card types are: `1`
```
cisacke: Update at some point to include the types you have so far

src/components/formfields/FormField.js
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Label is optional. If not provided, component will reorganize accordingly.
+   */
+  label: React.PropTypes.string,
+  /**
+   * For use with the `button-toggle` type.
+   *
+   */
+  toggleChoices: React.PropTypes.array,
+  selectOptions: React.PropTypes.array
```
cisacke: Missing `placeholder` (We should add a eslint rule to check for missing props since it's super annoying to keep in sync)

src/components/formfields/FormField.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+FormField.defaultProps = {
+  type: 'textfield',
+  forProp: 'first_name',
+  placeholder: 'Placeholder',
+  toggleChoices: ['Choice 1', 'Choice 2']
```
cisacke: We probably don't need `defaultProps` for anything that isn't required?

src/components/formfields/FormField.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export default FormField;
```
cisacke: Thinking about this a bit more & I think it might be better to split this up into separate components for each type. Currently there's no shared logic between the three types, and having them as separate components would improve readability (especially down the line as we add more types) & also eliminate unnecessary props that are specific to a type. We should discuss this when the three of us sync up at some point

src/components/formfields/FormField.js
drewdrewthis: @cisacke If you don't provide it, it just spits out an empty div. You're probably right, but it seemed not worth the time on this PR.

cisacke:
@drewdrewthis LGTM

drewdrewthis:
@cisacke Yes!

PR #13, opened 2017-04-10
Webpack reconfigure

jmcolella:
@drewdrewthis @cisacke


package.json
```diff
@@ -4,12 +4,15 @@
   "description": "React / Webpack / Babel framework for quick set up",
   "main": "index.js",
   "scripts": {
-    "build": "webpack --config ./webpack.prod.config.js --progress --colors",
+    "clean": "rm -rf ./dist/*",
+    "build": "webpack --config ./webpack.config.js --progress --colors",
+    "build:prod": "NODE_ENV='production' webpack --config ./webpack.config.js --progress --colors",
     "start": "node ./app.js",
     "dev": "webpack-dev-server --inline --hot",
```
cisacke: do we need this? if not, we could change `styleguide` to use `dev` in keeping with the other repos?

src/modules/cards/types/PercentageCard/index.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import TextLockup from '../../../../components/text-lockups/TextLockup';
```
cisacke: maybe use `src/components/...` instead of the relative path here

package.json
jmcolella: @cisacke good point. I've kept the dev server like this because it provides much better error messages than the styleguide build/server. Pretty dumb reason to keep it, but until I find a better way to log errors, I'm going to hold on to this

cisacke: gotcha, sounds good

src/modules/cards/types/PercentageCard/index.js
jmcolella: @cisacke thanks for pointing this out, added a resolve for the config

jmcolella:
@drewdrewthis @cisacke FYI merging.

PR #14, opened 2017-04-11
Configure file-loader for image assets

jmcolella:
@drewdrewthis @cisacke I updated the webpack config to use file-loader for building image assets


styleguide/index.html
```diff
@@ -4,8 +4,8 @@
 		<meta charset="utf-8">
 		<title>PolicyGenius React Component Library</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
-	</head>
+	<link href="assets/styles.css" rel="stylesheet"></head>
 	<body>
 		<div id="app"></div>
-	<script type="text/javascript" src="build/bundle.515c0810.js"></script></body>
+	<script type="text/javascript" src="build/bundle.fdd69a0d.js"></script></body>
```
j-clark: presumably this hash is going to  change with each build? in which case we should probably remove it so we don't have to update this file

jmcolella: Yeah, Drew updated the gitignore to exclude these files. it's in a different open PR

PR #15, opened 2017-04-11
Add Outline & Action Buttons

cisacke:
https://www.pivotaltracker.com/story/show/143344167

@jmcolella @drewdrewthis 👀  please if one of you gets the chance!


cisacke:
@drewdrewthis @jmcolella got confirmation from Natalie that renters doesn't need icons in the outline button (& as we already discussed, it doesn't appear in the flow we were just looking at) so I've removed everything icon related -- lmk if this looks good to merge

PR #18, opened 2017-04-12
Add image aside

jmcolella:
https://www.pivotaltracker.com/story/show/143293121

@drewdrewthis @cisacke 

This PR adds an Icon atom to render SVG inline and the new simple Image Aside molecule for the ticket.

Also updated linter to ignore bracket notation and cleaned up one errant prop type for Article List


src/atoms/icons/Icon.js
```diff
@@ -0,0 +1,38 @@
...
+    </div>
+  )
+}
+
+Icon.propTypes = {
+  /**
+   * name of icon to render;
+   * see `src/assets/images/index.js` for complete list
+   */
+  icon: React.PropTypes.string.isRequired,
```
cisacke: as we just discussed, let's do `import { PropTypes }`

src/molecules/asides/image-aside/index.js
```diff
@@ -0,0 +1,69 @@
...
+            icon={ icon }
+           wrapperClassName={ styles.icon } />
+        :
+          <img src={ icon } className={ styles.icon } />
+      }
+
+      <aside className={ styles.aside }>
+        <h5 className={ styles.header }>{ header }</h5>
+
+        { subheader ?
```
cisacke: you could do
```
{ subheader &&
  <p>...</p>
}
```
to avoid the empty binary

src/containers/App.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react';
```
cisacke: What's this file for / do we want it in the styleguide?

src/atoms/icons/Icon.js
jmcolella: @cisacke cool, will do

src/molecules/asides/image-aside/index.js
jmcolella: @cisacke yup yup sick 👍

src/containers/App.js
jmcolella: @cisacke so this file is the gateway to the DOM when running the dev server. it seems we have just moved away from this completely, so I just scrapped it. although the dev server has better error messages, we don't need this file as a 'testing ground' for errors and bloating the repo. Just need to figure out how to make the styleguide give us better errors....

cisacke: ahh ok, sounds good!

PR #19, opened 2017-04-12
Add step progress

cisacke:
https://www.pivotaltracker.com/story/show/143618135

Issue I was running into was trying to import a style directly into the `Readme.md` so that I could add an example wrapper around the component for demonstration purposes on small screens. I thought you could only require files in the config, but turns out you can add it to the markdown, so that's what I did. I propose that going forward, we can use `styleguide.scss` as a place to put styles that are being used in the documentation - but definitely open to ideas for a better implementation.

Decided to keep the old CSS and not use the `Icon` component because the old styleguide works very nicely out of the box & in the interest of getting the component done, I didn't want to re-write some fairly complex CSS.


jmcolella:
@cisacke can you export all of the new svgs in the icon `index.js`? Just would be helpful to have a full documentation of our icon library in that file

styleguide.config.js
```diff
@@ -23,6 +23,10 @@ module.exports = {
         {
           name: 'TextLockups',
           components: 'src/molecules/text-lockups/**/*.js'
+        },
+        {
+          name: 'StepProgress',
```
drewdrewthis: I suggest making an "Other" or "Misc" directory, so that we don't have to add a new section to the config file for every molecule.

styleguide/index.html
```diff
@@ -8,5 +8,5 @@
   <body>
     <div id="app"></div>
     <footer>We got a footer!</footer>
-  <script type="text/javascript" src="build/bundle.422eb0d6.js"></script></body>
+  <script type="text/javascript" src="build/bundle.77113a7e.js"></script></body>
```
drewdrewthis: This shouldn't be showing up. Make sure you've rebased your `.gitignore` and cleared your cache. https://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore

src/molecules/step-progress/step-progress.module.scss
```diff
@@ -0,0 +1,224 @@
+@import 'src/assets/stylesheets/global';
+
+$step-progress-color:                           $neutral-2;
```
drewdrewthis: Can we remove these `step-progress-` prefixes and not do this going forward?

src/assets/stylesheets/utilities/styleguide.scss
```diff
@@ -0,0 +1,10 @@
+@import 'src/assets/stylesheets/global';
+
+@media #{$small-only} {
+  .step-progress-wrapper {
+    padding: 16px;
+    background-color: $brand-1;
```
drewdrewthis: I think that this progress bar should have a transparent background -- but like we discussed, let's just leave it and move forward.

cisacke: 👍

PR #20, opened 2017-04-12
Update cards

drewdrewthis:


styleguide/index.html
```diff
@@ -7,6 +7,5 @@
   <link href="assets/styles.css" rel="stylesheet"></head>
   <body>
     <div id="app"></div>
-    <footer>We got a footer!</footer>
-  <script type="text/javascript" src="build/bundle.422eb0d6.js"></script></body>
+  <script type="text/javascript" src="build/bundle.556e3d32.js"></script></body>
```
cisacke: shouldn't be showing up?

src/atoms/TextComponents/textComponent.module.scss
```diff
@@ -0,0 +1,19 @@
+@import 'src/assets/stylesheets/global';
```
cisacke: so far we've been naming scss files `text-component.module.scss` etc - I don't mind either way but we should pick one and standardize

src/atoms/TextComponents/TextComponent.js
```diff
@@ -0,0 +1,56 @@
...
+    semibold,
+    light
+  } = props;
+
+  return React.createElement(
+    tag,
+    { className: classnames(
+      className,
+      styles['typography-' + type],
+      semibold ? styles.semibold : styles.regular,
```
cisacke: this would require you to explicitly pass `light={false}`, otherwise the binary below will add `styles.regular` and override the semibold styling. maybe there could be a `weight` prop which is optional and defaults to regular, with semibold & regular as the type choices

src/organisms/cards/AddCard/index.js
```diff
@@ -1,14 +1,42 @@
...
+    </div>
   )
 }
 
-AddCard.defaultProps = {
-  children: '+ Add New Beneficiary'
+function AddCard ( props ) {
+  return (
+    <div className={ styles.add }>
+      { renderContent(props) }
```
cisacke: would it make more sense to accept the icon name and text as props? otherwise every time we add a new `AddCard` type, we'll have to modify this component

styleguide/index.html
drewdrewthis: 👍

src/atoms/TextComponents/TextComponent.js
drewdrewthis: @cisacke Added a helper function.

src/organisms/cards/AddCard/index.js
drewdrewthis: @cisacke I was thinking that.. but there should only be so many of these. It does make adding new AddCards more complicated.. but it makes the API way easier.

cisacke: 👍  sounds good

cisacke:
@drewdrewthis LGTM!

PR #21, opened 2017-04-13
Add select card

jmcolella:
https://www.pivotaltracker.com/story/show/143664005

Adds select card and reorganizes Form Fields directory


jmcolella:
@drewdrewthis @cisacke 👀  please

styleguide/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
```
cisacke: do we want this file tracked?

src/utils/Fields/renderPlaceholder.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
```
cisacke: does it make sense to just treat this as a component and put it in `atoms`?

src/molecules/formfields/Select/SimpleSelectField/index.js
```diff
@@ -0,0 +1,102 @@
...
+        label &&
+        <label
+          className={ classnames( styles.label, { [styles.focused]: focused } ) }
+          htmlFor={ forProp }
+        >
+          { label }
+        </label>
+      }
+
+      <div className={ classnames( styles.selectbox ) }>
```
cisacke: there are a couple places in this pr where you don't need `classnames` - not a big deal but maybe remove them if they're unnecessary

src/molecules/formfields/Text/index.js
```diff
@@ -1,18 +1,25 @@
...
+    placeholder,
+    label
+  } = props;
 
   return (
     <div className={ classnames(className, styles.textfield) }>
-      <label
-        className={ styles.label }
-      >{ label }</label>
+      {
```
cisacke: nice!

src/organisms/cards/SelectCard/index.js
```diff
@@ -0,0 +1,95 @@
...
+        label={ label }
+        placeholder={ placeholder }
+        selectOptions={ selectOptions }
+        onChange={ onChange }
+        value={ value }
+      />
+
+      <footer className={ styles.footer }>
+        <div className={ styles['footer-link-wrapper'] }>
+          <a
```
cisacke: you might need to rebase but we have a link component now! also I wonder if eventually this footer could become a component?

src/organisms/forms/GeneralForm/Readme.md
```diff
@@ -7,11 +7,15 @@ _This is a form wrapper that can take components as children._
...
-        label='Home type'
-        placeholder='Select your home type'
+      <TextField
+        label='TextField Label'
+      />
+      <ToggleField
+        label='ToggleField Label'
+        toggleChoices={['Choice A', 'Choice B']}
+      />
+      <SimpleSelectField
```
cisacke: values/ labels aren't getting pulled through in this example

src/organisms/cards/SelectCard/index.js
jmcolella: yeah i agree, when we have time to refactor in the coming weeks/months

src/organisms/forms/GeneralForm/Readme.md
jmcolella: fixed thanks for the catch!

src/utils/Fields/renderPlaceholder.js
jmcolella: idk it's really just a helper for making a placeholder. honestly, I think we should move away from placeholders eventually, since react has good handling of default values for select fields

styleguide/index.html
jmcolella: no! i'll delete this

src/organisms/forms/GeneralForm/index.js
```diff
@@ -13,6 +13,7 @@ function GeneralForm ( props ) {
   } = props;
 
   const kids = React.Children.map(children, child => {
+    console.log('children', children);
```
cisacke: let's take this out

src/utils/Fields/renderOptGroup.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+export default function renderOptGroup( opt, optStyles ) {
+  return (
+    <optgroup label={ opt.group }>
```
cisacke: there's a weird thing where on initial load, this optgroup label is checked - it's not a blocker for merging this PR but we should fix eventually

src/molecules/formfields/Select/BasicSelectField/index.js
```diff
@@ -0,0 +1,106 @@
...
+    label,
+    selectOptions,
+    forProp,
+    focused,
+    onChange,
+    value
+  } = props;
+
+  return (
+    <div className={ classnames(className, styles.selectfield, { [styles.focused]: focused } ) }>
```
drewdrewthis: The `className` here needs to be the last param in the `classnames()` param list

src/molecules/formfields/Select/CondensedSelectField/condensed-select.module.scss
```diff
@@ -0,0 +1,35 @@
+@import 'src/assets/stylesheets/global';
+
+$condensed-label-color:  $neutral-2;
+$condensed-select-color: $brand-2;
```
drewdrewthis: I don't think we need these prefixes anymore.

src/molecules/formfields/Select/CondensedSelectField/index.js
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <div className={ classnames( styles.formfield, className ) }>
+      {
+        label &&
+        <label
+          className={ styles.label }
+          htmlFor={ forProp }
+        >
+          { label }
+        </label>
```
drewdrewthis: Awesome!

src/organisms/forms/GeneralForm/index.js
drewdrewthis: Please remove the `console.log`

src/organisms/forms/GeneralForm/Readme.md
```diff
@@ -7,22 +7,26 @@ _This is a form wrapper that can take components as children._
...
+        label='Select Field Label'
+        placeholder='Placeholder'
         selectOptions={[
-        ['Value 1', 'Option 1'],
-        ['Value 2', 'Option 2'],
-        ['Value 3', 'Option 3']
+          { value: 'Option 1' },
+          { value: 'Option 2' },
+          { value: 'Option 3' }
```
drewdrewthis: @jmcolella The reason I made this an array initially was so that the consumer could guarantee the order of the options, since it's my understanding that object key/value pairs don't guarantee that.

src/utils/Fields/renderPlaceholder.js
drewdrewthis: @jmcolella If you are consolidating the select fields, I imagine that this file will be deleted?

src/organisms/forms/GeneralForm/Readme.md
jmcolella: @drewdrewthis I think the `value` prop on the field itself should be the same name as the text it displays. this will make it much easier for end users to determine what option was actually selected. It's also a practice I've seen, particularly with handling `onChange` events with react

src/utils/Fields/renderPlaceholder.js
jmcolella: @drewdrewthis I'm hesitant to delete because it's really a utility function and doesn't belong within the file for the select field itself

src/organisms/forms/GeneralForm/Readme.md
drewdrewthis: @jmcolella Alright. But I feel like this is the same issue we discussed before about not giving people the option to create the mappings. What if the value needs to correspond to some other data object, but we want to display a pretty name for the user on the FE?

Also, this still doesn't guarantee order -- so, if we wanted a list of months in a certain order or like a list of years that are 100+ options long?

jmcolella:
@drewdrewthis any other comments besides the resolved array issue? would like to merge soon

PR #22, opened 2017-04-13
Add checkbox list

cisacke:
https://www.pivotaltracker.com/n/projects/1509328

on 2.5 in https://drive.google.com/drive/folders/0B_2yuT1AZna6TV95NUNwbmduLVk

### DIFF
--- a/src/molecules/formfields/CheckboxList.js
+++ b/src/molecules/formfields/CheckboxList.js
@@ -1,0 +1,45 @@
+import React, { PropTypes } from 'react';
+import styles from './formfields.module.scss';
+import classnames from 'classnames';
+
+function CheckboxList ( props ) {
+  const {
+    label,
+    focused,
+    data,
+    handleClick
+  } = props;
+
+  return (
+    <div className={ classnames(styles.checkboxList, { [styles.focused]: focused })}>
+      <label className={ styles.label }>{ label }</label>
+      <div className={ styles.content }>
+        {data.map((item, idx) => {
+          return (
+            <label className={ styles.checkbox } key={ idx } onClick={ () => handleClick( item ) }>
+              <input type='checkbox' defaultChecked={ item.checked } disabled={ item.disabled } />
+              <span className={ styles.checkboxLabel }>{ item.checkboxLabel }</span>
+            </label>
+          );
+        })}
+      </div>
+    </div>
+  );
+}
+
+CheckboxList.propTypes = {
+  label: PropTypes.string,
+  focused: PropTypes.bool,
+  data: PropTypes.arrayOf(
+    PropTypes.shape({
+      checkboxLabel: PropTypes.string.isRequired,
+      disabled: PropTypes.bool,
+      checked: PropTypes.bool
+    })
+  ),
+  handleClick: PropTypes.func.isRequired
+};
+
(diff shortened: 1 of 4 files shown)


src/molecules/formfields/CheckboxList.js
```diff
@@ -0,0 +1,46 @@
...
+    handleClick
+  } = props;
+
+  return (
+    <div className={ classnames(styles.checkboxList, { [styles.focused]: focused })}>
+      <label className={ styles.label }>{ label }</label>
+      <div className={ styles.content }>
+        {data.map((item, idx) => {
+          return (
+            <label className={ styles.checkbox } key={ idx } onClick={ () => handleClick( item.id ) }>
```
jmcolella: just curious why you inlined the `onClick` function? I feel like we should just pass a callback via props to give the end user more flexibility? at least for something like a checkbox, where the callback can register the event and the end user can manipulate that as they wish

src/molecules/formfields/CheckboxList.js
```diff
@@ -0,0 +1,46 @@
...
+  focused: PropTypes.bool,
+  data: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.number,
+      checkboxLabel: PropTypes.string.isRequired,
+      disabled: PropTypes.bool,
+      checked: PropTypes.bool
+    })
+  ),
+  handleClick: PropTypes.func
```
jmcolella: should this be required? seems essential for a checkbox

src/molecules/formfields/CheckboxList.js
jmcolella: eh maybe this is fine, idk

jmcolella: maybe just pass the `item` in rather than the id? seems very specific

src/molecules/formfields/formfields.module.scss
```diff
@@ -116,3 +129,123 @@ $section-padding: ru(.5) ru(.75);
...
   }
 }
+
+.checkboxList {
+  composes: formfield;
+
+  .content {
+    padding: ru(.75);
+
+    > * {
```
jmcolella: can we try to avoid using the wildcard selector?

src/molecules/formfields/CheckboxList.js
cisacke: for sure, i'll add it in!

src/molecules/formfields/formfields.module.scss
cisacke: yes, tbh I just blind copied this from styles, but I'll look into improving it

src/molecules/formfields/CheckboxList.js
cisacke: hmm yeah I was trying to think for a while how best to do this. The problem with passing back the item is that there's no guarantee of anything in the item being a unique identifier, although I guess that would be the responsibility of the end user to make sure there is something unique in the data... I will change to `item`, and I'm sure we'll get good feedback as devs start to implement stuff using the library

jmcolella: great, I think that's a good call

PR #23, opened 2017-04-13
Add linkwrapper and update cards

drewdrewthis:


styleguide/index.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
```
cisacke: Do we want this?

src/organisms/cards/AddCard/index.js
```diff
@@ -11,7 +11,7 @@ function renderContent( props ) {
   var txt, icon;
 
   if ( beneficiary ) {
-    txt = 'Add Beneficiary';
+    txt = 'Add person to policy';
```
cisacke: outside the scope of this PR but I'm still not sure about this component. I was looking at the designs & if you look at 2.4 in the flow, the text changes when you have already added a category, which means we'd need 2 types of component for each state. Like you mentioned last night, the trade off is having to specify `txt` and `icon` when you go to implement the component, but I think this is preferable to having to add all the conditionals for each possible state.

src/atoms/LinkWrapper/LinkWrapper.js
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <a
+      { ..._.omit(props, ['children', 'className']) }
+      className={ classnames(className, styles.link) }>
+      { children }
+    </a>
+  )
+}
+
+LinkWrapper.propTypes = {
```
cisacke: Different prop types needed here?

src/atoms/LinkWrapper/LinkWrapper.js
```diff
@@ -0,0 +1,50 @@
...
+
+function LinkWrapper ( props ) {
+  const {
+    className,
+    children,
+  } = props;
+
+  return (
+    <a
+      { ..._.omit(props, ['children', 'className']) }
```
cisacke: is the only prop you're using here `href`? if so you could declare that explicitly to help with readability

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "React / Webpack / Babel framework for quick set up",
   "main": "index.js",
   "scripts": {
+    "open": "open http://localhost:6060",
```
cisacke: 🙌

src/atoms/LinkWrapper/LinkWrapper.js
drewdrewthis: @cisacke It's passing through any standard anchor tag props, so the user could set `target`, `rel`, `href`, whatever.

src/atoms/LinkWrapper/LinkWrapper.js
drewdrewthis: wooops, good catch. Thanks @cisacke

src/organisms/cards/AddCard/index.js
drewdrewthis: Make sense.

PR #24, opened 2017-04-13
Fix overflow issue with image aside on mobile

jmcolella:
@drewdrewthis @cisacke 👀  please

fixes issue on mobile


src/molecules/asides/image-aside/index.js
```diff
@@ -21,7 +21,9 @@ function ImageAside ( props ) {
             icon={ icon }
             wrapperClassName={ styles.icon } />
         :
-          <img src={ icon } className={ styles.icon } />
+          <div className={ styles.icon }>
+            <img src={ icon } />
+          </div>
       }
 
       <aside className={ styles.aside }>
```
drewdrewthis: I think you should update this to use the `TextComponent` -- we don't want this h5 hardwired in, and we'd probably want the TextComponent to handle the typography

jmcolella: 👍

cisacke:
@jmcolella what's the issue this is fixing? I can't see anything wrong on master, but this also seems to work fine so LGTM! 🤷‍♀️

src/molecules/asides/ImageAside/index.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      <aside className={ styles.aside }>
+        <TextComponent
+          className={ styles.header }
+        >
+          { header }
+        </TextComponent>
+
+        { subheader && <TextComponent className={ styles.subheader }>{ subheader }</TextComponent> }
```
drewdrewthis: I suggest breaking this up into multiple lines, but not a blocker.

PR #27, opened 2017-04-13
Create navigator

drewdrewthis:


src/templates/Navigator/index.js
```diff
@@ -21,7 +50,11 @@ Navigator.propTypes = {
    * This prop will be add a new className to any inherent classNames
    * provided in the component's index.js file.
    */
-  className: React.PropTypes.string,
+  className: PropTypes.string,
+  /**
+   * Text to show up on left rail
+   */
+  txt: PropTypes.string
```
jmcolella: I think this prop name can be more descriptive

src/templates/Navigator/index.js
```diff
@@ -1,16 +1,45 @@
...
+          >
+            { txt }
+          </TextComponent>
+        }
+      </div>
+      <div className={ styles.main }>
+        <StepProgress
+          className={ styles['step-progress'] }
+          data={[
+            { active: false, inactive: false, icon: 'icon-calculator', label: 'Aenean', link: '#' },
```
jmcolella: a bit unrelated, but I think `StepProgress` should have `onClick` handlers for each item. Would make this component more flexible for the end user

src/templates/Navigator/navigator.module.scss
```diff
@@ -0,0 +1,32 @@
...
+.navigator {
+  // Place your component specific scss here.
+  background-color: $neutral-7;
+  height: 100%;
+  display: flex;
+}
+
+.left-rail {
+  align-self: flex-stretch;
+  width: ru(11);
```
drewdrewthis: Not really.. if they don't give us a grid with the layout, I don't think we should bother applying it.

src/templates/Navigator/index.js
drewdrewthis: Yeah, I agree UX wise, but that's something to discuss at the app level.

PR #30, opened 2017-04-14
Add prop types to form fields

cisacke:


src/molecules/formfields/SelectField/index.js
```diff
@@ -94,7 +96,10 @@ SelectField.propTypes = {
   /**
    * array containing objects for select value
    */
-  selectOptions: PropTypes.array.isRequired,
+  selectOptions: PropTypes.arrayOf(PropTypes.shape({
```
jmcolella: yeah awesome stuff here, I need to do this more often...

src/molecules/formfields/SelectField/Readme.md
```diff
@@ -4,9 +4,9 @@ Label Example:
       label='Select Field Label'
       placeholder='Placeholder'
       selectOptions={[
-        { value: 'Option 1' },
-        { value: 'Option 2' },
-        { value: 'Option 3' }
+        { label: 'Option 1', value: 1 },
```
jmcolella: smart. this takes care of Josh's feedback for the ticket, thanks!

jmcolella: @cisacke however, I think this changes up the `SelectCard` example, if I'm not mistaken, since it's wired to set a `value` state based on the value prop for the option, which then gets displayed to the user.

src/molecules/formfields/SelectField/index.js
```diff
@@ -31,10 +31,11 @@ function SelectField ( props ) {
       }
 
       <div className={ styles.selectbox }>
-        <select required
+        <select
+          required
           className={ styles.select }
           onChange={ onChange }
-          value={ value }
```
jmcolella: how come you switched this out in favor of `defaultValue`. it was my understanding that `defaultValue` is for uncontrolled components?

cisacke: mmm yes good catch I've switched it back, it should always be controlled

cisacke:
@jmcolella updated per your comments & made a fairly significant change -- since we're using redux form, the props that get passed down to form fields are going to be controlled by redux form, and they group them into `input`, `meta` and `...custom` so I've adjusted accordingly to plan for that. The history of why that is is [here](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.4), but going forward we should expect those props

src/molecules/formfields/SelectField/Readme.md
```diff
@@ -1,64 +1,50 @@
...
-      onChange={() => alert('Changing...')} 
-      focused />
+      focused
+    />
 
 Group Options Example:
 
 NOTE: `selectOptions` array must contain objects structured as follows:
 
 `{ group: 'groupName', options: [{value: 'optionName'}]}`.
```
jmcolella: just need to fix this documentation here based on the new structure

jmcolella:
@cisacke cool al LGTM. nice catch for redux form

PR #32, opened 2017-04-14
Add tooltip, update components

drewdrewthis:


src/pages/Navigator-Page-2.md
```diff
@@ -0,0 +1,39 @@
...
+      <Navigator leftRailText='Home sweet home' stepprogress={ false }>
+        <GeneralForm
+          formName='ExampleForm'
+          onSubmit={ () => alert('FormSubmitted!') }
+        >
+          <TextComponent tag='p' type={ 4 } semibold>
+            Great, just a few more questions.
+          </TextComponent>
+          <ToggleField
+            htmlFor=''
```
cisacke: `ToggleField` doesn't take `htmlFor` as a prop currently - do we want it to?

styleguide.config.js
```diff
@@ -50,6 +50,15 @@ module.exports = {
     {
       name: 'Templates',
       components: 'src/templates/**/*.js'
+    },
+    {
+      name: 'Pages',
```
cisacke: The first navigator page is under `Templates` - should we move it under Pages?

src/atoms/Tooltip/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+function Tooltip ( props ) {
+  const {
+    children,
+    onClick,
+    className
+  } = props;
+
+  return (
+    <div className={ classnames(className, styles['tooltip']) }>
```
cisacke: did we decide we want to add className to the end of the concatenation?

src/atoms/Tooltip/index.js
```diff
@@ -0,0 +1,38 @@
...
+      <Icon icon='tooltip' 
+        onClick={ onClick }
```
cisacke: `Icon` doesn't take an `onClick` prop - is it needed anywhere else in the component?

src/atoms/Tooltip/index.js
```diff
@@ -0,0 +1,38 @@
...
+Tooltip.propTypes = {
+  /**
+   * This prop will add a new className to any inherent classNames
+   * provided in the component's index.js file.
+   */
+  className: PropTypes.string,
+  onClick: PropTypes.func
+}
+
+Tooltip.defaultProps = {
```
cisacke: take this out if you don't need it

src/pages/Navigator-Page-2.md
drewdrewthis: @cisacke I think so.. even if it's not used in the implementation, it makes sense to have it

styleguide.config.js
drewdrewthis: @cisacke I thought about it, but we need an example for the navigator... anyway and I didn't want to just make a dummy example or have two of the same. What do you think?

src/atoms/Tooltip/index.js
drewdrewthis: good catch

src/pages/Navigator-Page-2.md
drewdrewthis: @cisacke Err... change of opinion. This isn't actually a form input. I think it was supposed to be a radio button implementation... but I don't think that's necessary since we're using React + Redux. I removed it.

src/atoms/Tooltip/index.js
```diff
@@ -0,0 +1,34 @@
...
+  const {
+    children,
+    onClick,
+    className
+  } = props;
+
+  return (
+    <div className={ classnames(styles['tooltip'], className) }>
+      <Icon icon='tooltip' 
+        onClick={ onClick }
```
cisacke: @drewdrewthis take all the `onClick` stuff out from this component and then this LGTM

styleguide.config.js
cisacke: ahh I see, ok yeah leave things as they are!

PR #33, opened 2017-04-16
Update button component

cisacke:
see rejection notes https://www.pivotaltracker.com/story/show/143344167

### DIFF
--- a/src/atoms/buttons/Button.js
+++ b/src/atoms/buttons/Button.js
@@ -4,38 +4,54 @@ import classnames from 'classnames';
 
 function Button ( props ) {
   const {
+    style,
     type,
     children,
     onClick,
     shake,
-    className
+    className,
+    disabled
   } = props;
 
   return (
-    <div
-      className={ classnames(className, styles[type], { [styles.shaking]: shake }) }
+    <button
+      className={ classnames({ [styles.shaking]: shake }, styles[style], className) }
       onClick={ onClick }
-    >{children}</div>
+      type={ type }
+      disabled={ disabled }
+    >
+      {children}
+    </button>
   )
 }
 
 Button.propTypes = {
   /**
-   * Possible button types are: `info, disabled, action, lowlight, actionDisabled, outline, or default (button)`
-   *
+   * Possible button styles are: `info`, `disabled`, `toggle`, `action`, `actionDisabled`, `lowlight`, `outline`, or `button` (default)
    */
-  type: PropTypes.string.isRequired,
-  onClick: PropTypes.func.isRequired,
+  style: PropTypes.string,
+  onClick: PropTypes.func,
   /**
    * Will append new classname to classSet
    */
-  className: React.PropTypes.string,
-  shake: PropTypes.bool
+  className: PropTypes.string,
+  /**
+   * Triggers shake animation if true
+   */
+  shake: PropTypes.bool,
+  /**
+   * Possible `type` attributes are: `submit`, `reset`, `button`, or `menu`
+   */
+  type: PropTypes.string,
+  /**
+   * Adds `disabled` attribute to button element if true
+   */
+  disabled: PropTypes.bool
 }
 
 Button.defaultProps = {
-  type: 'button',
-  onClick: () => alert('You clicked me!')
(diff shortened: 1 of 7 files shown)


src/atoms/buttons/Button.js
```diff
@@ -8,34 +8,44 @@ function Button ( props ) {
...
-    className
+    className,
+    buttonType,
+    disabled
   } = props;
 
   return (
-    <div
+    <button
       className={ classnames(className, styles[type], { [styles.shaking]: shake }) }
```
drewdrewthis: @cisacke I believe that the className prop should be last in the arg list to allow for overrides.

src/atoms/buttons/Button.js
```diff
@@ -8,34 +8,44 @@ function Button ( props ) {
...
    * Will append new classname to classSet
    */
   className: React.PropTypes.string,
-  shake: PropTypes.bool
+  shake: PropTypes.bool,
+  /**
+   * `Submit, reset, button, or menu`
+   */
+  buttonType: PropTypes.string,
+  disabled: PropTypes.bool
```
jmcolella: I think all `PropTypes` should have a description, please add one here

src/atoms/buttons/Button.js
```diff
@@ -8,34 +8,44 @@ function Button ( props ) {
...
 }
 
 Button.defaultProps = {
   type: 'button',
-  onClick: () => alert('You clicked me!')
+  buttonType: 'button'
```
drewdrewthis: @cisacke Why aren't we using `type` as the prop name? Seems like it's probably a good idea, and if so, we should probably standardize it.

src/atoms/buttons/Button.js
jmcolella: For the select card, Josh suggested we move towards explicit type props rather than having to pass in a type string. I came up with a decent solution for the select card, I think this is something we need to standardize

src/atoms/buttons/Button.js
```diff
@@ -8,34 +8,44 @@ function Button ( props ) {
...
-  onClick: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
   /**
    * Will append new classname to classSet
    */
   className: React.PropTypes.string,
-  shake: PropTypes.bool
+  shake: PropTypes.bool,
+  /**
+   * `Submit, reset, button, or menu`
```
drewdrewthis: @cisacke Perhaps we should explicitly state that these are the options available.

src/atoms/buttons/Button.js
cisacke: ahh ok, as in `<button action />` instead of `<button type="action" />`? where/ what's the solution for select card?

src/atoms/buttons/Button.js
cisacke: @drewdrewthis we are already using `type` in the way we are for all components but this is for the type attribute on a button element, so I needed to come up with a different name, hence `buttonType`. it is definitely confusing though and it seems like there's a move to deprecate the standard `type` prop in favor of explicit types, so if that's the case I'll change this to `type`.

src/atoms/buttons/Button.js
jmcolella: @cisacke whoops I meant `ImageAside`. https://github.com/policygenius/athenaeum/blob/master/src/molecules/asides/ImageAside/index.js

Look at `variants`. It's not perfect, but I think it's a pattern we can refine and use so all an end user needs to pass is `action` rather than `type='action'`

cisacke: @jmcolella gotcha, thanks - I will make the change for button but fwiw I'm not sure about this implementation. All we're doing with `type` at the moment is adding a new class in the sass file and exposing it as an option to the user. If we use explicit types, every time we want to add a new variation (and each component could have 10-15 variations), we'd have to declare it as a possible prop, add it to the classnames variant map, add the new class to the sass file, and expose to the user. It seems like a lot of unnecessary steps, when the trade off is asking the user to write `type="action"` instead of `action`. might be at least worth discussing before we decide to switch everything over

PR #34, opened 2017-04-17
Use better prop names for step progress

cisacke:
https://www.pivotaltracker.com/story/show/143618135


drewdrewthis:
@cisacke Great call. LGTM.

PR #35, opened 2017-04-17
Add layout spacers

jmcolella:


src/public/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import App from 'organisms/cards/SelectCard';
+import App from 'atoms/Layout';
```
drewdrewthis: @jmcolella What's this about? Should this be part of `.gitignore`?

src/atoms/Spacer/Readme.md
```diff
@@ -0,0 +1,44 @@
+Extra Small:
+
+    <div>
+    <div style={{ backgroundColor: '#b3b3b3', width: '100%', height: '100px'}}></div>
```
drewdrewthis: @jmcolella Consider creating a one-off example component and just giving it a dotted outline?

src/atoms/Layout/layout.module.scss
```diff
@@ -0,0 +1,115 @@
...
+
+      &:last-child {
+        margin-bottom: 0;
+      }
+    }
+  }
+}
+
+// Breakpoints
+.small {
```
drewdrewthis: @jmcolella Consider DRYing this up with functions/mixins

src/atoms/Layout/layout.module.scss
```diff
@@ -0,0 +1,115 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  flex-direction: column;
+
+  // Generate column classes
+  @for $i from 1 through 12 {
+    .column-#{$i} {
+      width: 100%;
+      margin: ru(.75) 0;
```
drewdrewthis: @jmcolella Seems like you can easily add in an `@include g-unit(#{$i});` here and have it just pump out exactly what we need.

src/public/index.js
jmcolella: @drewdrewthis I use this for better debugging. It should be ripped out

src/atoms/Layout/layout.module.scss
cisacke: @jmcolella I came across a sass tutorial a while back and he makes a grid system which I thought was really neat if you want to check it out --
http://callmenick.com/post/maintainable-responsive-web-design-with-sass
and the actual code:
http://codepen.io/callmenick/pen/dGjPOm?editors=1100

src/atoms/Spacer/Readme.md
jmcolella: @drewdrewthis I think having the blocks around the spacer really help to illustrate the spacing

src/public/index.js
jmcolella: @drewdrewthis I'm not ripping this out yet because the new dev webpack config that Josh made still uses this as the entry point

src/atoms/Spacer/index.js
```diff
@@ -0,0 +1,67 @@
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+import styles from './spacer.module.scss';
+
+function Spacer ( props ) {
+  const {
+    xSmall,
```
cisacke: I think this is a great example of when explicit prop types should be used instead of `type='...'` - really, a finite number of options instead of a potentially ever-growing set of component variations

src/atoms/Layout/index.js
```diff
@@ -0,0 +1,81 @@
...
+    verticalAlign
+  } = props;
+
+  let colsIdx = -1;
+
+  const kids = React.Children.map( children, ( child, idx ) => {
+    const length = !children.length ? 1 : children.length;
+    const childColsLength = childCols.length;
+
+    colsIdx = childColsLength == 1 || ( (idx + 1) > childColsLength && (idx + 1) % childColsLength === 1 ) ? -1 : colsIdx;
```
drewdrewthis: `childColsLength` should be an integer at this point, so you should be able to strictly compare (`===`)

drewdrewthis: This logic is hard to decipher. I'm imagining that it's supposed to loop over the ratios if you have more children than ratios provided? I personally think that something like this should have a little explanatory comment, but general PG policy is that if it needs a comment, you should refactor it to make it easier to understand.

src/atoms/Layout/index.js
```diff
@@ -0,0 +1,81 @@
...
+      </div>
+    </div>
+  )
+};
+
+Layout.propTypes = {
+  /**
+   * determines pixel breakpoint for columns and rows
+   */
+  breakpoint: PropTypes.string,
```
drewdrewthis: So.. since you can't tie "ratios" to breakpoints, then it's really just deciding when the rows switch up to full-width columns. It seems like this is something that can be handled by the parent / consumer - perhaps we should leave this out of the interface?

src/atoms/Layout/layout.module.scss
```diff
@@ -0,0 +1,96 @@
+@import 'src/assets/stylesheets/global';
+
+$gutter-width: 1.5rem;
+
+@mixin g-unit($cols: 1, $total-col-count: 12) {
```
drewdrewthis: @jmcolella Do we want devs to be able to use the `g-unit` without using the `Layout` component?

src/atoms/Spacer/Readme.md
```diff
@@ -0,0 +1,44 @@
+Extra Small:
```
drewdrewthis: @jmcolella What do you think about using a smart loop for these?

src/atoms/Spacer/index.js
drewdrewthis: 👍

src/atoms/Layout/index.js
jmcolella: @drewdrewthis thanks for helping me refactor

src/atoms/Layout/layout.module.scss
jmcolella: @drewdrewthis I think it would be confusing if we had 2 ways of creating a layout. It might be best to have this component be the source of truth for the grid

src/atoms/Layout/index.js
```diff
@@ -0,0 +1,81 @@
...
+    childCols,
+    breakpoint,
+    horizontalAlign,
+    verticalAlign
+  } = props;
+
+  let colsIdx = -1;
+
+  const kids = React.Children.map( children, ( child, idx ) => {
+    const length = !children.length ? 1 : children.length;
```
jmcolella: @drewdrewthis removed

src/atoms/Layout/index.js
jmcolella: @drewdrewthis I think it might be nice to have this functionality; it would certainly make changing the layout from row to column really easy without having to write extra CSS and pass in as a `className`. The default is to not break to columns anyway

src/atoms/Layout/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const kids = React.Children.map( children, ( child, idx ) => {
+    if ( !childCols[++colsIdx] ) {
+      colsIdx = 0;
+    }
+
+    return (
+      <div
+        className={ classnames(
+          styles[`column-${ childCols[colsIdx] || 12 }`]
```
drewdrewthis: I still think that the logic here is too opaque. The code is really nice tho -- so, personally, I'd like to see some comments here explaining what's going on. I think it's just a nice thing to do for any future engineers that might need to come in here.

jmcolella: @drewdrewthis OK will do.

PR #36, opened 2017-04-17
Add adelle-sans font

cisacke:
Not 100% sure that this is the right way to do this, so lmk if there's a better implementation


drewdrewthis:
@cisacke If it works, it LGTM.

PR #37, opened 2017-04-17
Setup build for NPM publish

j-clark:


src/entry.js
```diff
@@ -0,0 +1,18 @@
...
+  require.context('./organisms', true, /.*\.js/),
+  require.context('./templates', true, /.*\.js/),
+  // require.context('./pages', true, /.*\.js/),
+];
+
+module.exports = contexts.map((context) => fromPairs(context.keys().map((component) => {
+  const componentName = component.match(/.*\/(.*)\/index.js/)[1];
+
+  return [componentName, context(component).default];
+}))).reduce(mergeAll, {});
```
j-clark: This is a little complicated, but what we're doing here is dynamically identifying every file that is `**/*/Component/index.js` in the atoms, molecules, organisms and templates directories and exporting them at the top level

j-clark:
@policygenius/developers  hi this PR:

* sets up the publish process for this npm package, where all components are exported at the top level
* renames some files to match the ComponentName/index.js naming convention
* extracts the common webapack config
* cleans up dependencies that aren't needed or were only dev dependencies

package.json
```diff
@@ -1,38 +1,31 @@
...
-    "build": "webpack --config ./webpack.config.js --progress --colors",
-    "build:prod": "NODE_ENV='production' webpack --config ./webpack.config.js --progress --colors",
-    "start": "node ./app.js",
-    "dev": "webpack-dev-server --inline --hot",
+    "build": "./node_modules/webpack/bin/webpack.js --config ./webpack.config.dev.js --progress --colors",
+    "build:prod": "./node_modules/webpack/bin/webpack.js --config ./webpack.config.prod.js --progress --colors",
+    "publish": "yarn run publish",
+    "start": "yarn run styleguide:build && yarn run styleguide",
     "test": "karma start",
     "stylin": "yarn run styleguide:build && yarn run styleguide",
```
drewdrewthis: @j-clark This `stylin` does the exact same command as `start`. Maybe it's better to use the `start` convention.. perhaps change `stylin` to `yarn run start`.

src/entry.js
```diff
@@ -0,0 +1,18 @@
...
+const assign = require('lodash/assign');
+
+const mergeAll = (accum, current) => assign(accum, current);
+
+const contexts = [
+  require.context('./atoms', true, /.*\.js/),
+  require.context('./molecules', true, /.*\.js/),
+  require.context('./organisms', true, /.*\.js/),
+  require.context('./templates', true, /.*\.js/),
+  // require.context('./pages', true, /.*\.js/),
```
drewdrewthis: @j-clark I'm assuming that this is here just in case we want to add it later? If so, maybe add a little comment or even remove it, since it's highly unlikely that we'll need to export whole compositions of PG page examples.

src/entry.js
drewdrewthis: Neat.

.gitignore
```diff
@@ -46,3 +46,4 @@ styleguide
 # Ignore OSX system files
 .DS_Store
 
+lib/
```
drewdrewthis: @j-clark Probably want to add the `yarn.lock` to the `.gitignore`?

package.json
j-clark: ah, yeah. we might want to just stick with start since that is a pretty typical convention

.gitignore
j-clark: so the purpose of the yarn lockfile is to ensure that everyone gets exactly the same version of all dependencies, including transitive dependencies. so like a Gemfile.lock, we need to commit it to version control in order to accomplish that

src/entry.js
j-clark: sure i can remove it

.gitignore
drewdrewthis: 👍

package.json
drewdrewthis: @j-clark If you don't mind, leave me the `stylin` -- it puts a smile on my face in the morning.

PR #39, opened 2017-04-17
Fixes small styling tweaks to complete Navigator ticket

drewdrewthis:
https://www.pivotaltracker.com/story/show/143675961

This doesn't fix everything -- this ticket is a running work in progress, so this can be merged after review.

### DIFF
--- a/src/atoms/Icon/icons.module.scss
+++ b/src/atoms/Icon/icons.module.scss
@@ -2,11 +2,19 @@
 
 .icon-wrapper {
   position: relative;
-  width: ru(1);
-  height: ru(1);
+  width: ru(3);
+  height: ru(3);
+
+  .SVGInline-svg {
+    position: relative;
+    width: 100%;
+    height: 100%;
+  }
 
   svg {
     position: absolute;
+    left: 0;
+    top: 0;
     width: 100%;
     height: 100%;
   }
--- a/src/atoms/Icon/icons.module.scss
+++ b/src/atoms/Icon/icons.module.scss
@@ -1,0 +1,13 @@
+@import 'src/assets/stylesheets/global';
+
+.icon-wrapper {
+  position: relative;
+  width: ru(1);
+  height: ru(1);
+
+  svg {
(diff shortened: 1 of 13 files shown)


src/atoms/icons/icons.module.scss
```diff
@@ -0,0 +1,15 @@
+@import 'src/assets/stylesheets/global';
```
cisacke: might have been a weird merge with Josh's PR but this has ended up in `/icons` instead of `/Icon`

src/atoms/Icon/index.js
```diff
@@ -13,7 +14,7 @@ function Icon ( props ) {
 
   return (
     <div
-      className={ classnames(className, wrapperClassName) }
+      className={ classnames(styles['icon-wrapper'], className, wrapperClassName) }
```
cisacke: adding `icon-wrapper` class to this looks like it makes the icons disappear in the styleguide examples

src/atoms/Icon/icons.module.scss
```diff
@@ -0,0 +1,19 @@
+@import 'src/assets/stylesheets/global';
+
+.icon-wrapper {
+  width: 100%;
```
cisacke: @drewdrewthis there's duplicate `width` and `height` in here - also this styling is making the pg icon in the navigator/ navigator 2 pages tiny

drewdrewthis: @cisacke You're right about the duplicates -- I'll add new sizes in the navigator, but the idea is to basically set a standard `ru(1)` sq dimension here that can be resized by whatever component uses the icon. FYI this was rebuilt because SVGs can proportionally fill there containers.

cisacke: @drewdrewthis gotcha - ok remove the duplicate styles & fix the examples and then merge

drewdrewthis:
@cisacke Added an icon map. Had @jmcolella do a quick CR before merging.

PR #42, opened 2017-04-18
Fix the styleguide

j-clark:
@policygenius/developers


webpack.config.base.js
```diff
@@ -19,8 +19,8 @@ module.exports = (options) => ({
   context: path.resolve(__dirname, 'src'),
   resolve: {
     modules: [
-      path.resolve('./src'),
-      path.resolve('./node_modules')
+      path.resolve(__dirname, 'src'),
+      'node_modules'
```
j-clark: didn't really look into why, but this is the change that was causing the issue, so i reverted it to its previous state

jmcolella: I must have changed this because of an issue with using absolute paths for the import statements. However, if this works fine , then sounds good.

PR #46, opened 2017-04-19
Navigator Page 1 (Con't)

drewdrewthis:
https://www.pivotaltracker.com/story/show/143675961

### DIFF
--- a/src/atoms/Layout/Col/index.js
+++ b/src/atoms/Layout/Col/index.js
@@ -15,12 +15,12 @@ function Col(props) {
     xxLargeCols,
     fullwidth,
     flex,
-    colStyle,
+    style,
   } = props;
 
   return (
     <div
-      style={colStyle}
+      style={style}
       className={classnames(
         { [styles[`col-sm-${smallCols}`]]: smallCols },
         { [styles[`col-md-${mediumCols}`]]: mediumCols },
@@ -74,7 +74,7 @@ Col.propTypes = {
    * adds additional styles to the column as a React styles object  ``
    * - For useful positional styles, [Checkout this Guide to Flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)
    */
-  colStyle: PropTypes.object
+  style: PropTypes.object
 };
 
 
--- a/src/templates/Navigator/index.js
+++ b/src/templates/Navigator/index.js
@@ -1,11 +1,55 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import Layout from 'atoms/Layout';
+import Col from 'atoms/Layout/Col';
 import Icon from 'atoms/Icon';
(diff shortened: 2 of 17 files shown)


src/atoms/Layout/Col/index.js
```diff
@@ -15,12 +15,12 @@ function Col(props) {
     xxLargeCols,
     fullwidth,
     flex,
-    colStyle,
+    style,
```
drewdrewthis: @patrickkim This forces the user to use a `Col` component explicitly if they way to give it `style` or `className`, since these are reserved words. We should update the doc.

patrickkim: Makes sense. I'll update the docs.

drewdrewthis:
@policygenius/front-end Please CR! Thanks. A bunch of updates in here -- the Navigator is pretty much ready to go as a layout. It needs a Footer component (template), but that's it, I believe.

src/templates/Navigator/index.js
```diff
@@ -1,12 +1,56 @@
...
+import Layout from 'atoms/Layout';
+import Col from 'atoms/Layout/Col';
 import Icon from 'atoms/Icon';
 import TextComponent from 'atoms/TextComponent';
 import StepProgress from 'molecules/StepProgress';
 import ContactCard from 'organisms/cards/ContactCard';
 import styles from './navigator.module.scss';
 
+
+function renderStepProgress() {
```
cisacke: I think having stepProgress in the navigator component is going to be difficult to use because we'll need to control it from the app side - we can probably just take it out here & have the end user implement it

src/templates/Navigator/index.js
```diff
@@ -1,12 +1,56 @@
...
+          inactive: true,
+          icon: 'icon-quotes',
+          label: 'Etiam',
+          link: '#' }
+      ]}
+    />
+  );
+}
+
+function renderContactCard() {
```
cisacke: Looks like the right rail will need tweaking as well because depending on where you're at in the flow, you might have the map or the card with the policy details in addition to the contact card. Could we just expose the right rail as its own component inside the Navigator so that it's purely presentational?

src/templates/Navigator/navigator.module.scss
```diff
@@ -3,37 +3,140 @@
...
   display: flex;
 }
 
-.left-rail {
-  align-self: flex-stretch;
-  width: ru(11);
-  padding: ru(1.5);
+.layout {
+  align-items: stretch;
+  max-width: $page-max-width;
```
cisacke: I think we want this to take up the full width of the page?

cisacke:
@drewdrewthis ^^ few things noted above but this is looking good!

src/templates/Navigator/navigator.module.scss
drewdrewthis: Nope... the layout gets the max width, but the navigator itself is fullwidth.

PR #48, opened 2017-04-20
Test setup

cisacke:
Super light testing setup using Jest - pretty much works out of the box. The only thing I had to configure was the css files - because we use CSS modules, we need to use [identity-obj-proxy](https://www.npmjs.com/package/identity-obj-proxy) which mocks classNames nicely.


drewdrewthis:
@cisacke This LGTM. @jmcolella had a comment about using Chai?

drewdrewthis:
@cisacke If we're going to use this for all the tests, can you please update the plopfile / templater?

cisacke:
@drewdrewthis yup good call -- added to the plopfile
@jmcolella added Chai 🍵

.babelrc
```diff
@@ -0,0 +1,4 @@
+{
```
jmcolella: is this file necessary? these presets are already set in the webpack config

src/atoms/Button/__tests__/button.spec.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Button from '../index';
```
jmcolella: this relative import should go after the absolute imports; the linter will complain

plop_templates/test_template.txt
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import sinon from 'sinon';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
```
jmcolella: maybe add the import statement for the component? `import {{properCase name }} from {{type}}/{{properCase name}}`...or maybe not all components will have this specific path? just a thought

src/atoms/Button/__tests__/button.spec.js
```diff
@@ -0,0 +1,67 @@
...
+    };
+  });
+
+  it('is a <button> tag', () => {
+    const wrapper = shallow(<Button />);
+    expect(wrapper.type()).to.equal('button');
+  });
+
+  it('adds the style prop as a className', () => {
+    const props = assign({}, defaultProps, {style: 'info'});
```
jmcolella: slight wrench here...this test will fail with my PR. `style` is a reserved word in React and must have an object as its value. The linter complains about this and I agree, so I changed this prop value to `kind`. We need to once and for all have a team meeting about how to handle variant prop names

plop_templates/test_template.txt
patrickkim: Yes, I think this would help.

src/atoms/Button/__tests__/button.spec.js
patrickkim: make it happen @jmcolella

cisacke: @jmcolella ahh ok - I'll hold off merging this until your PR is ready, and then I'll switch over to using `kind` or whatever we decide to use

plop_templates/test_template.txt
cisacke: 👍  good call - I think it's safe to assume they'll have the same relative path

.babelrc
cisacke: @jmcolella yeah I'm not sure to be honest - when I remove the `.babelrc` the js doesn't compile when the tests run but I don't know why. Although I did discover that you can put the presets in `package.json` instead of creating a .babelrc so I've done that instead

src/atoms/Button/__tests__/button.spec.js
patrickkim: > variant

cisacke:
@patrickkim resolved conflicts & changed button to use variant so tests will pass - lmk if this looks good to merge!

PR #51, opened 2017-04-21
Fix sass setup

drewdrewthis:
The "base" styles were being added to the main styles.css for each component. This extracts it into a base.scss which will be required before all of the component styles. It also sets a standard of using scss partials correctly.


src/entry.js
```diff
@@ -1,3 +1,5 @@
+require('src/assets/stylesheets/base.css');
```
drewdrewthis: @j-clark Is this correct for your config, or should it be added to the  contexts?

cisacke:
LGTM

src/entry.js
```diff
@@ -1,3 +1,5 @@
+require('./assets/stylesheets/base.scss');
```
j-clark: to your question about whether or not this is right, the way to verify would be to build the project `npm run build:publish` and see what's in the `lib/` directory.

i think this will probably produce this file as an artifact, but am i going to have to include 2 separate stylesheets in order to use this in another project? that would be my only question

PR #52, opened 2017-04-21
Add prop-type package, Remove uses of React.PropType

patrickkim:


patrickkim:
Need CR. Pretty straight forward 👀  @cisacke @drewdrewthis @jmcolella

cisacke:
@patrickkim LGTM

PR #55, opened 2017-04-24
Fix webpack config for publishing

j-clark:


j-clark:
@policygenius/developers

PR #56, opened 2017-04-24
Add dotenv to store clientside keys

patrickkim:
use `process.env[’SECRET KEY FROM ENV’]` to use secret


j-clark:
out curiosity, what's this going to be used for — as in what keys does the RCL need?

patrickkim:
I figure it would be useful to add a Google Maps API so that you can preview a map component. I'm open to other suggestions.

drewdrewthis:
@patrickkim LGTM if you think we'll use it.

patrickkim:
Don't need this after all. The consumer should handle their respective api keys.

PR #57, opened 2017-04-24
Clean up low hanging pending items.

patrickkim:
- Fix missing key prop in children.
- Button uses `variant` prop over `kind.
- Cleaned up helper function a tad.
- Fix test_template's beforeEach.


patrickkim:
:eyes: CR @drewdrewthis

PR #60, opened 2017-04-24
set up toggle field to work with redux form

cisacke:
This sets up the toggle field to work with redux form - I don't think this is the most ideal/ long term solution because it tightly couples the component with redux form, and ideally we would want some kind of wrapper component that handles redux form specific interactions. But until we decide on that, this will work for the intended purpose, and the component can still be used on its own by passing in an `onChange` callback instead of `onClick`

### DIFF
--- a/src/molecules/formfields/ToggleField/Readme.md
+++ b/src/molecules/formfields/ToggleField/Readme.md
@@ -2,7 +2,14 @@
   <ToggleField
     htmlFor=''
     label='Toggle Choices'
-    toggleChoices={[ 'Choice 1', 'Choice 2' ]}
+    toggleChoices={[
+      { label: 'Choice 1', value: 1 },
+      { label: 'Choice 2', value: 2 }
+    ]}
     tooltipMessage='You can choose either of these awesome options!'
+    input={{
+      value: state.value,
+      onChange: (value) => setState({ value })
+    }}
   />
 ```
--- a/src/molecules/formfields/ToggleField/index.js
+++ b/src/molecules/formfields/ToggleField/index.js
@@ -3,15 +3,17 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Tooltip from 'atoms/Tooltip';
 import Button from 'atoms/Button';
+import omit from 'lodash/omit';
+
 import styles from '../shared/formfields.module.scss';
 
 function ToggleField( props ) {
   const {
     label,
     className,
-    onClick,
     tooltipMessage,
-    toggleChoices
+    toggleChoices,
+    input
   } = props;
 
   return (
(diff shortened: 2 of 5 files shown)


src/molecules/formfields/ToggleField/index.js
```diff
@@ -35,13 +35,14 @@ function ToggleField( props ) {
...
           <Button
-            variant='toggle'
+            variant={input.value === choice.value ? 'toggle-active' : 'toggle'}
             key={`button-toggle-btn-${idx}`}
             className={styles.button}
-            onClick={onClick}
-          >{ choice }</Button>
-          )
-        }
+            onClick={() => input.onChange(choice.value)}
```
jmcolella: I'm confused by your PR comment; it seems to use this component w/o ReduxForm, one would still need to pass in a `input` prop in order for this `onClick` handler to work. Your comment states that one would only need to pass in `onChange`. Am I missing something or did I take your comment too literally?

src/molecules/formfields/ToggleField/index.js
```diff
@@ -57,17 +58,31 @@ ToggleField.propTypes = {
...
+    })
+  ),
   /**
    * The `tooltopMessage` can be anything that can be rendered:
    * `numbers`, `strings`, `elements` or an `array` (or fragment) containing these types.
    */
-  tooltipMessage: PropTypes.node
+  tooltipMessage: PropTypes.node,
+  /**
+   * @ignore
```
jmcolella: should this be ignored? See comment above regarding `onChange` usage w/o Redux Form

src/molecules/formfields/ToggleField/Readme.md
```diff
@@ -2,7 +2,14 @@
...
     label='Toggle Choices'
-    toggleChoices={[ 'Choice 1', 'Choice 2' ]}
+    toggleChoices={[
+      {label: 'Choice 1', value: 1},
+      {label: 'Choice 2', value: 2}
+    ]}
     tooltipMessage='You can choose either of these awesome options!'
+    input={{
+      value: state.value,
+      onChange: (value) => setState({value})
```
jmcolella: I didn't know you could do this with `setState`. Do you have a link to docs/an article about it. I can't find anything currently

cisacke: https://github.com/styleguidist/react-styleguidist/blob/master/docs/Documenting.md under 'Writing Code Examples' - I didn't know this either until last night, could be a nicer way of writing stateful examples rather than having to write a dummy component wrapper

j-clark:
for this to work with redux-form, don't we just need to have a specific API exposed? e.g. we need value, onChange, onBlur, onFocus, etc. props

i think if the props inside input were flattened at the top-level, we'd have a long-term solution that would work

src/molecules/formfields/ToggleField/index.js
cisacke: @jmcolella  yeah sorry my comment was confusing - I meant that the way this is currently implemented, you can pass an `onClick` handler for the buttons, but that the changes I'm proposing would require you to pass `onChange` (and a value) inside of the `input` object to achieve the same functionality which is kinda annoying because we have to behave as if we're redux form in these cases

src/molecules/formfields/ToggleField/index.js
cisacke: hmm yeah the reason I ignored it was to prevent the defaultProp from showing the `onChange` function I wrote - I didn't want to require it, although maybe I should because it's the only way the component will work

cisacke:
@j-clark ok yeah that makes sense - I flattened all of the `input` props into `Button` and then passed input specific props into the `button` tag. I think only the `onClick` function needs to be written in order to trigger `onChange` since we don't have an input in the toggle field component, & the other props onFocus, onBlur etc. seem to work fine with the button

PR #61, opened 2017-04-27
Simplify stepProgress code.

patrickkim:
Get some CR 👀 @cisacke @drewdrewthis


cisacke:
@patrickkim  ^^

drewdrewthis:
@cisacke @patrickkim In the styles, `pointer-events` is set to `none` -- I believe that we actually don't want to give the user that ability.. you can check with Jason Fromm though, he's the Renter's PM. We can probably handle this issue in another ticket/PR though

patrickkim:
Carry over from the old styles. I'm going to remove them. We'll see if we need this in the future.

patrickkim:
currently its a `href` but show me a use case in where it needs to change and it will be done.

PR #62, opened 2017-04-27
Nav components

drewdrewthis:


drewdrewthis:
@patrickkim @cisacke I've extracted the main parts of the navigator and the user can now pass the stepprogress data directly through.

src/templates/Navigator/MainNavigatorComponent/Readme.md
```diff
@@ -0,0 +1,5 @@
+MainNavigatorComponent Example:
+
+    <MainNavigatorComponent>
```
j-clark: just a comment, that we can probably avoid adding `Component` to the names of our components. they're all components, so this suffix doesn't add any new information

drewdrewthis: @j-clark Yeah... I started off doing that, but then I thought it would be confusing, since it's really only to be used with the Navigator as a wrapping component... what do you think about `NavigatorMain` and `NavigatorSidebar`?

drewdrewthis: Because we don't want to use a general name that might conflict with another component.

j-clark: yeah, i see the complication.  a couple suggestions for how the API could look

```javascript
import { Navigator } from 'athenaeum';
import { NavigatorBody } from 'athenaeum';
import { NavigatorSidebar } from 'athenaeum';

export default () => (
  <Navigator>
    <NavigatorBody>
      { /* stuff here */ }
    </NavigatorBody>
    <NavigatorSidebar>
      { /* stuff here */ }
    </NavigatorSidebar>
  </Navigator>
);
```

or perhaps one step better is to group the related components like so:

```javascript
import { Navigator } from 'athenaeum';

export default () => (
  <Navigator>
    <Navigator.Body>
      { /* stuff here */ }
    </Navigator.Body>
    <Navigator.Sidebar>
      { /* stuff here */ }
    </Navigator.Sidebar>
  </Navigator>
);
```

whatchu thank?

drewdrewthis: The latter is what I was going for, I just didn't know we could do that.

drewdrewthis:
@patrickkim --> LGTM

PR #65, opened 2017-04-27
Toggle field: coerce everything to string before comparing

cisacke:
@policygenius/front-end v quick review 👀  pls!


jmcolella:
LGTM mergin for you

PR #67, opened 2017-04-27
Stacked cards are here!

patrickkim:


src/organisms/cards/StackedCard/__tests__/stacked_card.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  // beforeEach(() => {
+  //   defaultProps = {
+  //     // Add default props here
+  //   };
+  // });
+
+  it('is a <StackedCard>', () => {
+    const wrapper = shallow(<StackedCard />);
+
+    expect(wrapper).to.be.ok;
```
cisacke: are you just testing here that the component renders something? if that's the case I think
```
expect(wrapper.length).to.equal(1);
```
would be a better assertion because wrapper is always truthy, even for components that render falsey values

src/organisms/cards/StackedCard/__tests__/stacked_card.spec.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+// import sinon from 'sinon';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import StackedCard from '../index';
+
+describe('StackedCard', () => {
+  // let defaultProps;
```
cisacke: take this & the sinon import out if you don't need them

plop_templates/test_template.txt
```diff
@@ -1,20 +1,22 @@
...
-import sinon from 'sinon';
+//import sinon from 'sinon';
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
 
 import {{properCase name}} from '../index';
 
 describe('{{properCase name}}', () => {
-  let defaultProps;
+  // let defaultProps;
```
cisacke: I think we probably want to keep this in the test_template?

src/organisms/cards/ContactCard/index.js
```diff
@@ -17,30 +18,27 @@ function ContactCard( props ) {
...
-          onClick={chatClick}
-        >
-          <Icon
-            icon='chat'
-            className={styles.icon}
-          />{ chatText }
-        </a>
-      </li>
-    </ul>
+    <StackedCard className={classnames(styles.contact, className)}>
```
cisacke: nice! 🙌

src/organisms/cards/StackedCard/index.js
```diff
@@ -0,0 +1,30 @@
...
+const stackChildren = children => React.Children.map(children, stackChild);
+
+function StackedCard(props) {
+  const {
+    children,
+    className
+  } = props;
+
+  return (
+    <ul className={classnames(styles['stacked'], className)}>
```
cisacke: one other thing you can test for is that props.className gets appended to this div - this might be overkill so lmk what you think!

plop_templates/test_template.txt
patrickkim: I agree, which is why I commented them out. I want the first test template to not immediately fail linting.

src/organisms/cards/StackedCard/__tests__/stacked_card.spec.js
patrickkim: It doesn't test anything really just a default test but you bring up a good point.

plop_templates/test_template.txt
```diff
@@ -1,20 +1,22 @@
...
+  //
+  // beforeEach(() => {
+  //   defaultProps = {
+  //     // Add default props here
+  //   };
+  // });
```
patrickkim: Left the comments on the template because they are important but want to pass linters on first pass.

PR #70, opened 2017-04-28
Build separate stylesheet for military.com

j-clark:


j-clark:
@policygenius/developers 

This PR allows us to build a new stylesheet for our military.com partnership. We are able to add any overrides as follows

* there is a new directory: variants/military_dot_com
* for every scss file in there, we will find it's base stylesheet counterpart and append the affiliate overrides
  * example: if you define a file variants/military_dot_com/stylesheets/base_styles/_colors.scss and define a new `$brand-2`, that color definition will be appended to the contents of src/stylesheets/base_styles/_colors.scss when building a new file military_dot_com_styles.css

This would support adding any new affiliate in the future by adding a directory of overrides in variants/my_new_affiliate

bin/build_assets
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const fsExtra = require('fs-extra');
+const walk = require('klaw-sync');
+const exec = require('shelljs').exec;
+
+const variants = fs.readdirSync('variants');
```
trevornelson: Is there a specific reason you're using fs and fs-extra? I **think** any fs methods are also in fs-extra. If they need to be syncronous, fs-extra should be able to do that as well.

j-clark: didn't realize that!

j-clark:
all right, it's all using fs-extra now.

definitely want the RCL folks to see the import changes at least

patrickkim:
Late to the game here. LGTM, although I'm curious as to how we might be able to implement this in wild. @josh @drewdrewthis can we sync up to discuss packaging css.

PR #71, opened 2017-04-28
Price discount header

patrickkim:


src/molecules/PriceDiscountHeader/index.js
```diff
@@ -0,0 +1,51 @@
...
+  label: PropTypes.string,
+  /**
+   *  previous amount. There are no checks to see that its lower than the current price.
+   */
+  prevAmount: PropTypes.number,
+
+};
+
+PriceDiscountHeader.defaultProps = {
+  amount: 0,
```
cisacke: If `amount` is required, we probably don't need default props

src/molecules/PriceDiscountHeader/price_discount_header.module.scss
```diff
@@ -0,0 +1,37 @@
...
+  padding: ru(.5) ru(1);
+
+  background: $brand-3;
+  color: $neutral-8;
+
+  line-height: 1;
+  text-align: center;
+}
+
+// small, sup { color: $debug; }
```
cisacke: still need this?

src/molecules/PriceDiscountHeader/__tests__/price_discount_header.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    const wrapper = shallow(<PriceDiscountHeader />);
+
+    expect(wrapper.length).to.equal(1);
+    expect(wrapper.find('.label')).to.have.length(0);
+    expect(wrapper.find('.prev-amount')).to.have.length(0);
+  });
+
+  it('shows label if prop.label is set', () => {
+    const wrapper = shallow(<PriceDiscountHeader amount={100} label='stuff' />);
+
```
cisacke: I suggest that we don't test for `.label` and `.prev-amount` to be length 0 in the previous test and to test them in their respective `it` blocks instead. You could omit them for the initial render, test they are 0, then set props to include them & test for length 1. example: https://github.com/policygenius/athenaeum/blob/master/src/atoms/Button/__tests__/button.spec.js#L35

src/molecules/PriceDiscountHeader/price_discount_header.module.scss
patrickkim: no we don't

src/molecules/PriceDiscountHeader/__tests__/price_discount_header.spec.js
patrickkim: all valid points.

PR #73, opened 2017-05-01
Add DataRow

patrickkim:


src/molecules/DataRow/data_row.module.scss
```diff
@@ -0,0 +1,39 @@
+@import 'assets/stylesheets/global';
+
+
+.data-row {
+  // Place your component specific scss here.
```
jmcolella: Can you remove this comment?

src/molecules/DataRow/index.js
```diff
@@ -0,0 +1,63 @@
...
+    PropTypes.string,
+    PropTypes.number,
+  ]).isRequired,
+  /**
+   * optional helper text describing this data.
+   */
+  description: PropTypes.string,
+};
+
+DataRow.defaultProps = {
```
jmcolella: Do we need this if there aren't any default Props?

src/molecules/DataRow/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+import styles from './data_row.module.scss';
+
+function DataRow(props) {
+  const {
+    className,
+    variant,
+    label,
+    value,
+    description,
```
jmcolella: stray comma here

src/molecules/DataRow/index.js
```diff
@@ -0,0 +1,63 @@
...
+  } = props;
+
+  return (
+    <div className={classnames(styles['data-row'], styles[variant], className)}>
+      <span className={classnames(styles['label'])}>
+        { label }
+        { description && <Tooltip className={styles['tip']}>{ description }</Tooltip> }
```
jmcolella: Looking at the screenshot on Clubhouse (https://app.clubhouse.io/policygenius/story/1861/datarow), it seems like a `label` can have a `tooltip` and it doesn't seem like this implementation will allow for that. Am I thinking about this wrong?

patrickkim: It would if it's passed in as a description prop

PR #83, opened 2017-05-02
Fix layout

drewdrewthis:


src/atoms/Layout/index.js
```diff
@@ -5,17 +5,37 @@ import _ from 'lodash';
 import styles from './layout.module.scss';
 import Col from './Col';
 
-function Layout( props ) {
+function processChild(child, layoutProps) {
+
+  if (child.type === Col) {
+    const colProps = Object.assign(
```
jmcolella: I would add `{}` as the first argument for `Object.assign` so that it returns a new object rather than completely overwriting `layoutProps`

src/atoms/Layout/index.js
```diff
@@ -5,17 +5,37 @@ import _ from 'lodash';
...
+      layoutProps,
+      child.props
+    );
+
+    return React.cloneElement(child, colProps);
+  }
+
+  const colProps = Object.assign(
```
jmcolella: See suggestion above

PR #85, opened 2017-05-02
Add Datefield Component, Add ComponentWrapper to replace FieldWrappers

patrickkim:


src/molecules/formfields/DateField/__tests__/date_field.spec.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import DateField from 'molecules/formfields/DateField';
+
+describe('DateField', () => {
+  xit('is a <DateField>', () => {
```
drewdrewthis: @patrickkim Why is this disabled?

src/molecules/formfields/DateField/index.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import dateUtils from 'utils/dateUtils';
+import renderOption from 'utils/Fields/renderOption';
+
+import styles from 'molecules/formfields/shared/formfields.module.scss';
+
+const renderMonthOption = (month, idx) => renderOption({
```
drewdrewthis: @patrickkim It would be nice great to have better test coverage on all of these functions.

src/molecules/formfields/DateField/index.js
```diff
@@ -0,0 +1,93 @@
...
+  label: PropTypes.string,
+  /**
+   * input object contains any props to be passed directly: value.
+   */
+  input: PropTypes.object.isRequired
+};
+
+DateField.defaultProps = {
+  // Place any default props here.
+  label: 'Date',
```
drewdrewthis: Might be nice to have this component default to today's date.

src/molecules/formfields/SelectField/Readme.md
```diff
@@ -1,33 +1,60 @@
...
+        value: 1 },
+      { label: 'Option 2',
+        value: 2 },
+      { label: 'Option 3',
+        value: 3 }
+    ];
 
     <SelectFieldExample
-      placeholder='Placholder'
+      placeholder='Placeholder'
```
drewdrewthis: Nice :)

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -163,8 +167,31 @@ $checkbox-border-color: $neutral-5;
   }
 }
 
-// Select Field - types
 
+//
+.datefield {
+  composes: formfield;
```
drewdrewthis: Since this file is getting so long, it might make sense break these all up into their own files -- if you do this, we would need to stop using `composes` and use a Sass placeholder instead to prevent class duplication.

patrickkim: Agree. We should avoid using composes in the future. Our sass should be able to compile without using CSS Loader.

src/molecules/formfields/DateField/index.js
patrickkim: True. Added a story to flesh those out

src/molecules/formfields/DateField/index.js
patrickkim: More work than its worth.

PR #86, opened 2017-05-03
Nav layout form fix

drewdrewthis:


drewdrewthis:
@policygenius/front-end Please CR thx :)

src/templates/Navigator/index.js
```diff
@@ -32,6 +32,9 @@ function separateParts(children) {
 }
 
 function renderPartial(part, data) {
+  // Will not render if data is undefined
+  // To render without data, pass in true
+  if (!data) return undefined;
```
patrickkim: Is this guard to prevent stepProgress component from throwing up?

drewdrewthis: @patrickkim haha, throwing up? This is to prevent a partial from adding empty space to the DOM when it doesn't have what it needs to render its component.

drewdrewthis:
@patrickkim The tests have passed ☝️

PR #87, opened 2017-05-03
Add color classes to Link Wrapper

jmcolella:


drewdrewthis:
@jmcolella This LGTM, but if we're going to actually be using this in a component like this, instead of just the example,  we should consider dropping the variables entirely and just use the map throughout the code base. This would eliminate the redundancy and possibility of a disconnect.

src/atoms/LinkWrapper/link_wrapper.module.scss
```diff
@@ -1,6 +1,18 @@
 @import 'assets/stylesheets/global';
 
+@mixin createColors( $type ) {
+  @for $i from 1 through 6 {
```
patrickkim: I think you can do `@each $key, $value in $map` to do this.

patrickkim:
LGTM

jmcolella:
@drewdrewthis can you take a once over on this?

Updates all color references to use `color-map` function
Updates bug with selectfield focused state border

src/assets/stylesheets/base_styles/_colors.scss
```diff
@@ -1,36 +1,42 @@
...
 /* Black. From dark to light. */
-$neutral-1: #000;
-$neutral-2: #4d4d4d;
-$neutral-3: #999;
-$neutral-4: #b3b3b3;
-$neutral-5: #ccc;
-$neutral-6: #f2f2f2;
-$neutral-7: #fbfbfb;
-$neutral-8: #fff;
+neutral-1: #000,
```
drewdrewthis: @jmcolella Is there a reason these aren't indented?

src/atoms/LinkWrapper/link_wrapper.module.scss
drewdrewthis: @jmcolella ☝️  You should be able to dynamically iterate over these now instead of explicitly stating a number, no?

PR #89, opened 2017-05-03
Meta props are available on all input/select fields

patrickkim:
Supports focus state via redux-forms

### DIFF
--- a/src/molecules/formfields/CheckBoxField/index.js
+++ b/src/molecules/formfields/CheckBoxField/index.js
@@ -60,8 +60,7 @@ CheckboxList.propTypes = {
       checked: PropTypes.bool
     })
   ),
-  handleClick: PropTypes.func.isRequired
+  handleClick: PropTypes.func.isRequired,
 };
 
 export default CheckboxList;
-
--- a/src/molecules/formfields/CheckBoxField/index.js
+++ b/src/molecules/formfields/CheckBoxField/index.js
@@ -2,23 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import toggleClass from 'utils/ReactUtils/toggleClass';
-
 import styles from '../shared/formfields.module.scss';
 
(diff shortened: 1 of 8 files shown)


drewdrewthis:
@patrickkim Please squash.

patrickkim:
Going to test out this squash and merge.

src/molecules/formfields/CheckBoxField/index.js
```diff
@@ -1,18 +1,24 @@
...
   const {
     label,
     focused,
     data,
-    handleClick
+    handleClick,
+    meta,
   } = props;
 
+  const focusedClass = toggleClass(focused || (meta && meta.active), styles['focused']);
```
cisacke: I think you can probably remove the `focused` prop entirely because we can still programmatically set the focused state through the `meta` object (if we're not using redux-form)

src/molecules/formfields/DateField/index.js
```diff
@@ -32,12 +33,13 @@ function DateField(props) {
...
     label,
     input,
+    meta,
   } = props;
 
   const dateParts = dateUtils.dateStrObj(input.value);
 
   return (
-    <div className={classnames(styles['datefield'], className)}>
+    <div className={classnames(styles['datefield'], toggleClass(meta && meta.active, styles['focused']), className)}>
```
cisacke: instead of extracting this into a helper function, you could use `classnames`:
```
classnames(styles['datefield'], {[styles.focused]: meta && meta.active}, className)
```

src/molecules/formfields/SelectField/index.js
```diff
@@ -21,19 +22,22 @@ function SelectField( props ) {
...
     type,
     input,
     tooltip,
-    defaultValue
+    focused,
+    defaultValue,
+    meta,
   } = props;
 
+  const focusedClass = toggleClass(focused || (meta && meta.active), styles['focused']);
```
cisacke: same as above, I think you can remove `focused`

PR #90, opened 2017-05-03
Small UX fixes

drewdrewthis:


src/molecules/formfields/ToggleField/Readme.md
```diff
@@ -8,8 +8,8 @@
     ]}
     tooltipMessage='You can choose either of these awesome options!'
     input={{
-      value: state.value,
-      onChange: (value) => setState({ value })
+      value: 'state.value',
```
cisacke: Patrick has a fix for the ToggleField error in https://github.com/policygenius/athenaeum/pull/89 so we don't need to change the Readme here

drewdrewthis: @cisacke @patrickkim The Readme still doesn't have any variables in it -- I don't see how his fix addresses this?

drewdrewthis: Anyway, I've removed it. If it's still an issue we can fix it in another PR

PR #91, opened 2017-05-03
Fix select field values

jmcolella:


jmcolella:
@cisacke @drewdrewthis @patrickkim 

Fixes mixup between `label` and `value` for SelectFields. This need to be CR'd soon as it is blocking renters tests/implementation.

cisacke:
@jmcolella LGTM

PR #93, opened 2017-05-04
Roll up utils & write tests

patrickkim:


src/molecules/formfields/SelectField/Readme.md
```diff
@@ -21,7 +21,7 @@ Label Example:
 No Label Example:
 
 ```example
-    const SelectFieldExample = require('utils/example_wrappers/ComponentWrapper')(SelectField);
+    const SelectFieldExample = require('utils/ComponentWrapper')(SelectField);
```
drewdrewthis: @patrickkim We probably want to keep these labeled as a wrapper for Examples specifically.

src/utils/fieldUtils/__tests__/renderOptGroup.spec.js
```diff
@@ -0,0 +1,34 @@
...
+import renderOptGroup from 'utils/fieldUtils/renderOptGroup';
+
+describe('renderOptGroup Util', () => {
+
+  it('renders an option group', () => {
+    const optGroup = {
+      reactKey: 'reactChildKey',
+      group: 'gruppe',
+      options: [
+        { label: 'labelle', value: 1 },
```
drewdrewthis: haha, cute.

src/molecules/formfields/SelectField/Readme.md
patrickkim: It's a pretty generic my thinking was that would be the only example wrapper. TBH didn't give it much thought to the name. If we have more and it gets confusing, I'll rename them.

PR #95, opened 2017-05-04
Add onClick to icon 

cisacke:
specifically for add form - we want to have an onClick handler for the `x` icon in the corner

### DIFF
--- a/src/atoms/Icon/index.js
+++ b/src/atoms/Icon/index.js
@@ -10,11 +10,13 @@ function Icon( props ) {
   const icon = icons[props.icon];
   const {
     className,
+    onClick
   } = props;
 
   return (
     <div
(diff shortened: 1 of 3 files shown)


src/atoms/Icon/index.js
```diff
@@ -31,7 +33,11 @@ Icon.propTypes = {
   /**
    * The icon you'd like to display.
    */
-  icon: PropTypes.string
+  icon: PropTypes.string,
+  /**
+   * click handler for the wrapper div around the svg
```
drewdrewthis: I know we have yet to standardize this, but should write these with the first letter capitalized?

cisacke:
@drewdrewthis made those changes!

PR #98, opened 2017-05-04
The new standard, DateFields are all input fields.

patrickkim:


drewdrewthis:
@patrickkim Okay, LGTM -- but why the change?

patrickkim:
Design says this is to be the new pattern. I got some questions on how we want to handle a grouping of fields like this.

PR #103, opened 2017-05-08
Add Modal component

jmcolella:
@drewdrewthis CR please. thanks!

Adds Modal component
Adds spacing prop for Layout component

### DIFF
--- a/src/assets/images/close.svg
+++ b/src/assets/images/close.svg
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="31px" height="30px" viewBox="0 0 31 30" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 43.2 (39069) - http://www.bohemiancoding.com/sketch -->
+    <title>Group 6</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="PPC-Calculator-Modal" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="square">
+        <g id="Artboard" transform="translate(-729.000000, -21.000000)" stroke="#4D4D4D" stroke-width="2">
+            <g id="Group-6" transform="translate(731.000000, 22.000000)">
+                <path d="M0.728,0.728 L27.4367713,27.4367713" id="Line"></path>
+                <path d="M0.728,0.728 L27.4367713,27.4367713" id="Line-Copy-2" transform="translate(13.832000, 13.832000) scale(-1, 1) translate(-13.832000, -13.832000) "></path>
+            </g>
+        </g>
+    </g>
+</svg>
--- a/src/assets/stylesheets/base_styles/_colors.scss
+++ b/src/assets/stylesheets/base_styles/_colors.scss
@@ -31,7 +31,10 @@ $colors-map: (
   action-3: #fab350,
   action-4: #fbc67c,
   action-5: #fdd9a7,
-  action-6: #feecd3
+  action-6: #feecd3,
+
+  /* Modal overlay background */
+  transparent-neutral-2: rgba(77,77,77,0.8)
 );
 
 @function color($type) {
--- a/src/assets/stylesheets/utilities/_mixins.scss
+++ b/src/assets/stylesheets/utilities/_mixins.scss
@@ -34,8 +34,8 @@
     $column-width: (100% / $per-row);
 
     @media ( min-width: $min-width * $per-row ) {
-      min-width: calc(#{$column-width} - #{$gutter});
       flex: 1 0 calc(#{$column-width} - #{$gutter});
+      min-width: calc(#{$column-width} - #{$gutter});
     }
 
     @for $i from 2 through $per-row {
@@ -43,10 +43,9 @@
       $column-width: (100% / $i);
 
       @media ( min-width: $screen-width ) {
-        max-width: calc(#{$column-width} - #{$gutter});
         flex: 1 0 calc(#{$column-width} - #{$gutter});
+        max-width: calc(#{$column-width} - #{$gutter});
       }
     }
   }
 }
-
--- a/src/molecules/Modal/__tests__/modal.spec.js
+++ b/src/molecules/Modal/__tests__/modal.spec.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import sinon from 'sinon';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import Modal from 'molecules/Modal';
+
+describe('<Modal />', () => {
+  let wrapper;
+  const callback = sinon.spy();
+
+  beforeEach(() => {
+    wrapper = shallow(<Modal contentLabel='Test Modal' />);
+  });
+
+  it('renders', () => {
+    expect(wrapper.length).to.equal(1);
+  });
+
+  it('passes props to ReactModal', () => {
+    wrapper.setProps({
+      onAfterOpen: callback,
+      onRequestClose: callback,
+      isOpen: false,
+      variant: 'simple'
+    });
+
+    expect(wrapper.props().className).to.contain('simple');
+    expect(wrapper.props().onAfterOpen).to.equal(callback);
+    expect(wrapper.props().onRequestClose).to.equal(callback);
+    expect(wrapper.props().isOpen).to.equal(false);
+  });
+
+  it('sets additional className prop', () => {
+    wrapper.setProps({
+      className: 'test'
+    });
+
+    expect(wrapper.props().className).to.contain('test');
+  });
+
+  it('calls onRequestClose callback when .close is clicked', () => {
+    wrapper.setProps({
+      onRequestClose: callback,
+    });
+
+    wrapper.find('.close').simulate('click');
+
+    expect(callback.called).to.equal(true);
+  });
+});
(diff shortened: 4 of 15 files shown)


src/assets/stylesheets/utilities/_mixins.scss
```diff
@@ -34,19 +34,49 @@
...
     }
 
     @for $i from 2 through $per-row {
       $screen-width: ($min-width * $i) + ($gutter * $i);
       $column-width: (100% / $i);
 
       @media ( min-width: $screen-width ) {
-        max-width: calc(#{$column-width} - #{$gutter});
         flex: 1 0 calc(#{$column-width} - #{$gutter});
+        max-width: calc(#{$column-width} - #{$gutter});
```
drewdrewthis: @jmcolella Why did you flip these? j/w

jmcolella: hahah I was getting a VIM linter error so I switched them...really dumb since it doesn't matter

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+
+.header {
+  padding: 1.875rem ru(2);
+  background-color: color('neutral-6');
+  border-bottom: 1px solid color('neutral-5');
+  color: color('neutral-2');
+}
+
+.close {
+  @include close(rem-calc(30px), rem-calc(2px), color('neutral-2'));
```
drewdrewthis: @jmcolella We probably don't have a ton of different 'X's -- this is really an icon. IMO, you should make this its own component, or a variant of the `Icon` component. Otherwise, every time we want to use this X, we'll have to look up the mixin and look up each of the params..

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.body {
+  font-family: $font-family-sans-serif;
+  padding: ru(2);
+  background-color: color('neutral-8');
+}
+
+.overlay {
+  background-color: rgba(77,77,77,0.8);
```
drewdrewthis: @jmcolella I'm surprised this passed the linter. It should probably be set as a variable / part of the color map

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+  z-index: 999;
```
drewdrewthis: @jmcolella I think that setting an arbitrarily high number for z-index is kind of bad practice. It really only needs to be the highest, which is often just `z-index: 1`

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+  transition: opacity 0.25s ease-in-out, background 0.25s ease-out;
```
drewdrewthis: @jmcolella We should also probably standardize our transitions.

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.dialog {
+  position: relative;
+  z-index: 9999;
+  max-width: 100%;
+  margin: auto;
+  border-radius: ru(.25);
+  border: 1px solid color('neutral-5');
+  font-family: 'adelle-sans', Helvetica, Arial, sans-serif;
```
drewdrewthis: @jmcolella You shouldn't be setting this directly. Please use the text mixin to maintain our standards / branding.

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,72 @@
...
+  overflow: hidden;
+}
+
+.header {
+  padding: 1.875rem ru(2);
```
drewdrewthis: @jmcolella This should be `ru()`

src/molecules/Modal/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * variant for modal
+   */
+  variant: PropTypes.string,
+
+  /**
+   * overlay variant for modal
+   */
+  overlay: PropTypes.string
```
drewdrewthis: @jmcolella Is this not something that should be a bool?

src/molecules/Modal/index.js
```diff
@@ -0,0 +1,101 @@
...
+    overlay
+} = props;
+
+  return (
+    <ReactModal
+      isOpen={isOpen}
+      onAfterOpen={onAfterOpen}
+      onRequestClose={onRequestClose}
+      contentLabel={contentLabel}
+      className={classnames(styles['modal'], styles[variant], className)}
```
drewdrewthis: @jmcolella Dot notation? `styles.modal`

src/molecules/Modal/__tests__/modal.spec.js
```diff
@@ -0,0 +1,22 @@
...
+import Modal from 'molecules/Modal';
+
+describe('<Modal />', () => {
+  // let defaultProps;
+  //
+  // beforeEach(() => {
+  //   defaultProps = {
+  //     // Add default props here
+  //   };
+  // });
```
drewdrewthis: @jmcolella ☝️ Please remove the commented out bits.

drewdrewthis: Also, you might consider adding a test to make sure that all of the props get correctly passed to the ReactModal

src/molecules/Modal/modal.module.scss
jmcolella: 👍

src/molecules/Modal/index.js
jmcolella: I left it as a string in case we want different styles of the overlay. I'm not sure if we should do that anyway

src/molecules/Modal/modal.module.scss
patrickkim: are you appending it to root?

jmcolella: It's all good; I copy/pasted from `life-web` JS styles. This actually maps to `neutral-2`

src/molecules/Modal/index.js
jmcolella: OK got rid of this

src/molecules/Modal/modal.module.scss
jmcolella: @drewdrewthis I set default values for the mixin. Will create a new `Icon` once natalie provides a SVG for the X

jmcolella:
@drewdrewthis can I get a CR please?

Adds `close` SVG icon
Refactor per your notes

src/assets/images/close.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
drewdrewthis: We should probably do some research into what we can/should remove from these svg files

src/assets/stylesheets/base_styles/_colors.scss
```diff
@@ -31,7 +31,10 @@ $colors-map: (
   action-3: #fab350,
   action-4: #fbc67c,
   action-5: #fdd9a7,
-  action-6: #feecd3
+  action-6: #feecd3,
+
+  /* Modal overlay background */
+  transparent-neutral-2: rgba(77,77,77,0.8)
```
drewdrewthis: 👍

src/molecules/Modal/modal.module.scss
```diff
@@ -0,0 +1,87 @@
...
+$modal-body-background-color:               color('neutral-8');
+$modal-body-color:                          color('neutral-2');
+$modal-body-subtitle-color:                 color('neutral-3');
+$modal-tab-border-color:                    color('neutral-5');
+$modal-tab-background-color:                color('neutral-7');
+$modal-tab-active-background-color:         color('neutral-8');
+$modal-tab-active-color:                    color('neutral-1');
+$robust-card-mobile-top-panel:              color('neutral-7');
+$robust-card-mobile-top-panel-border-color: color('neutral-5');
+$robust-card-mobile-top-panel-name-color:   color('neutral-3');
```
drewdrewthis: @jmcolella It doesn't look like any of these variables are used in this file? Please remove them if not. Also, I know this was a pattern from the old library, but I don't think we need to list everything as a variable up top like this. We really only need to use variables if two or more properties are sharing the same values and only when it makes sense, imo.

src/molecules/Modal/index.js
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * variant for modal
+   */
+  variant: PropTypes.string,
+};
+
+Modal.defaultProps = {
+  variant: 'simple',
+  headerWeight: 'semibold'
```
drewdrewthis: @jmcolella Why do you think you need to expose these right away, instead of waiting for there to be a need?

src/molecules/Modal/modal.module.scss
patrickkim: ⬆️ 👍

src/molecules/Modal/index.js
jmcolella: @drewdrewthis We already have a need.
Compare:
1. https://app.zeplin.io/project/58e6498f59f26ca94d348983/screen/5908b06faeb11aa4758b30c1

2. https://app.zeplin.io/project/58e6498f59f26ca94d348983/screen/5904cd5484b6b58d207eabaf

src/molecules/Modal/modal.module.scss
jmcolella: @drewdrewthis my bad, I forgot to delete

src/molecules/Modal/index.js
drewdrewthis: @jmcolella Please double check with Natalie that we actually need these two options. It would be better for everyone if there was just one standard.

drewdrewthis:
@jmcolella After those last two things (checking with Natalie and the variables), it LGTM. Excellent work.

PR #106, opened 2017-05-09
Checkout layout

drewdrewthis:


drewdrewthis:
@patrickkim Please squash your commits.

src/templates/Checkout/index.js
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <div className={classnames(styles['checkout'], className)}>
+      <Layout
+        className={styles.layout}
+        fullwidth
+      >
+        <Col className={styles.head}>
+          <Layout
+            mediumCols={[ 4 ]}
+            style={{ justifyContent: 'space-between' }}
```
drewdrewthis: @patrickkim There's no real reason not to do this via a class in the CSS, is there?

src/templates/Checkout/index.js
```diff
@@ -0,0 +1,82 @@
...
+import styles from './checkout.module.scss';
+
+function CheckOut(props) {
+  const {
+    children,
+    className,
+  } = props;
+
+  const parts = separatePartials(children);
+  const renderPartial = partialRenderer(CheckOutPartials);
```
drewdrewthis: @patrickkim Consider renaming `partialRenderer` into something like `createPartialRenderer()`

src/templates/Navigator/index.js
```diff
@@ -5,41 +5,11 @@ import classnames from 'classnames';
 import Layout from 'atoms/Layout';
 import Col from 'atoms/Layout/Col';
 import StepProgress from 'molecules/StepProgress';
+import { Partial, separatePartials, partialRenderer } from 'atoms/Partial';
```
drewdrewthis: @patrickkim Consider putting `Partial` into a new directory of helpers or something, and maybe even packaging them up into an RCL library like `lodash`... Making this an Atom I think is confusing.

drewdrewthis:
@patrickkim LGTM with some comments.. but tbh I think this needs some more thought. The level of abstraction and of splitting everything into so many files creates a complicated set up for each template that I think can be simplified / standardized.

src/templates/Checkout/index.js
patrickkim: nope artifact

src/templates/Navigator/index.js
patrickkim: Moved to `/helper`. Agree the interface could be tighter, made a chore to do that along with unifying/ writing more tests.

PR #108, opened 2017-05-10
Fix svgs

drewdrewthis:


bin/create_svg_manifest.js
```diff
@@ -0,0 +1,32 @@
+const glob = require('glob');
+const _ = require('lodash');
+const fs = require('fs'); // eslint-disable-line
```
patrickkim: the linter doesn't let you fs?

bin/create_svg_manifest.js
```diff
@@ -0,0 +1,32 @@
...
+function reducer(acc, curr) {
+  const camel = curr.split('.')[0];
+
+  return Object.assign({}, acc, {
+    [_.camelCase(camel)]: `./${curr}`
+  });
+}
+
+function generateManifest(err, matches) {
+  const manifest = matches.reduce(reducer, {});
```
patrickkim: 👍

src/assets/images/index.js
```diff
@@ -1,22 +1,5 @@
...
-export const pgLogoOrange = require('!svg-inline-loader!assets/images/pg-logo-orange.svg');
-export const phone = require('!svg-inline-loader!assets/images/phone.svg');
-export const plus = require('!svg-inline-loader!assets/images/plus.svg');
-export const quotes = require('!svg-inline-loader!assets/images/quotes.svg');
-export const rclLogo = require('!svg-inline-loader!assets/images/rcl-logo.svg');
-export const tooltip = require('!svg-inline-loader!assets/images/question-mark.svg');
-export const xIcon = require('!svg-inline-loader!assets/images/x-icon.svg');
+const manifest = require('./manifest.json');
+
+Object.keys(manifest).forEach((key) => {
```
patrickkim: You could use your reducer pattern here as well. Also the preferred standard for umd is to have a single export at the end of the file `module.export = {obj}`

src/assets/images/index.js
```diff
@@ -1,22 +1,5 @@
-export const application = require('!svg-inline-loader!assets/images/application.svg');
-export const calculator = require('!svg-inline-loader!assets/images/calculator.svg');
-export const cart = require('!svg-inline-loader!assets/images/cart.svg');
```
patrickkim: I believe this svg is making the checkout test fail.

bin/create_svg_manifest.js
j-clark: can you think of a more descriptive name for the reducer? since it's the function argument to `.reduce`, we know it's a reducer, so the method name doesn't give us an additional information

bin/create_svg_manifest.js
```diff
@@ -0,0 +1,32 @@
+const glob = require('glob');
+const _ = require('lodash');
+const fs = require('fs'); // eslint-disable-line
+
+function reducer(acc, curr) {
+  const camel = curr.split('.')[0];
+
+  return Object.assign({}, acc, {
+    [_.camelCase(camel)]: `./${curr}`
```
j-clark: same comment with camel. i think this name is actually misleading. since you're camelizing it, it's probably not already camelized. is it perhaps the `fileName`?

bin/create_svg_manifest.js
drewdrewthis: It was throwing errors because of my local npm or something, I think. I can revisit it today tho

bin/create_svg_manifest.js
drewdrewthis: :+1:

drewdrewthis:
@patrickkim @j-clark Holy bananas, batman. This was a bit insane. But it works now. @patrickkim I'm not sure what you did that was so special in the Checkout component that required such a complicated mock, but yeah. ✌️

src/templates/Navigator/__tests__/navigator.spec.js
```diff
@@ -10,6 +10,12 @@ import TextComponent from 'atoms/TextComponent';
 import Navigator from 'templates/Navigator';
 import NavigatorPartials from 'templates/Navigator/NavigatorPartials';
 
+/* eslint-disable no-undef */
```
patrickkim: `/* eslint-disable no-undef, no-shadow` You can disable multiple rules.

drewdrewthis: ahh, comma! of course

src/templates/CheckOut/__tests__/checkout.spec.js
```diff
@@ -3,7 +3,20 @@ import { expect } from 'chai';
 import { shallow, render } from 'enzyme';
 
 import CheckOut from 'templates/CheckOut';
-// import styles from 'templates/CheckOut/checkout.scss';
+
+/* eslint-disable */
+jest.mock('atoms/Icon', () => {
```
patrickkim: Does this work? If so, I'm glad you discovered this, might be better alternative to 'rewire'

drewdrewthis: Yeah, it works.

drewdrewthis: I thought we decided not to use rewire already when we first started on navigator

PR #110, opened 2017-05-10
Split checkboxList into wrapper and item

cisacke:
This will be necessary for implementation using redux-form, so we can wrap each individual checkbox in a `<Field>` component


cisacke:
@patrickkim @jmcolella @drewdrewthis  👀  pls

PR #112, opened 2017-05-10
Fix external React issue with ReactDOM and adjust modal

jmcolella:
@cisacke @drewdrewthis @patrickkim 

Fixes duplicate React issue in Renters (excludes ReactDOM from build)
Fixes alignment issue with Modal header

Please CR ASAP as the duplicate React fix is urgent


cisacke:
@jmcolella LGTM

PR #116, opened 2017-05-10
Credit cards

patrickkim:
This is going to have to change later on but would like it to get it out there.

### DIFF
--- a/src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
+++ b/src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import CreditCardField from 'molecules/formfields/CreditCardField';
+
+/* eslint-disable */
+jest.mock('atoms/Icon', () => {
+  const React = require('react');
+
+  class Icon extends React.Component {
+    render() {
+      return <div />;
+    }
+  }
+
+  return <div />;
+});
+/* eslint-enable */
+
+describe('<CreditCardField />', () => {
+
+  xit('renders', () => {
+    const wrapper = shallow(<CreditCardField />);
+
+    expect(wrapper.length).to.equal(1);
+  });
+});
--- a/src/molecules/formfields/CreditCardField/credit_card_field.module.scss
+++ b/src/molecules/formfields/CreditCardField/credit_card_field.module.scss
@@ -1,0 +1,47 @@
+@import 'assets/stylesheets/global';
+
+$form_border: 1px solid color('neutral-5');
+$padding: ru(.5);
+
+.credit-card {
+  // Place your component specific scss here.
+
+  &-header {
+    display: flex;
+    align-items: center;
+    padding-right: $padding;
+  }
+
+  &-lock {
+    width: ru(.75);
+    height: ru(.75);
+    margin-left: auto;
+  }
+
+  &-line1 {
+    position: relative;
+    border-bottom: $form_border;
+  }
+
+  &-number {
+    width: 100%;
+  }
+
+  &-logo {
+    position: absolute;
+    top: 0;
+    right: $padding;
+    height: 100%;
+    pointer-events: none;
+  }
+
+  &-line2 {
+    display: flex;
+  }
+
+  &-exp, &-sec { width: 50%; }
+
+  &-exp {
+    border-right: $form_border;
+  }
+}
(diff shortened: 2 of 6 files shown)


src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+// import sinon from 'sinon';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import CreditCardField from 'molecules/formfields/CreditCardField';
+
+describe('<CreditCardField />', () => {
```
cisacke: can we remove this?

src/molecules/formfields/CreditCardField/index.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const baseClass = [
+    styles['formfield'],
+    styles['credit-card'],
+    { [styles.focused]: meta && meta.active },
+    className
+  ];
+
+  return (
+    <div className={classnames(...baseClass)}>
```
cisacke: this looks great, but I think another implementation would be to do this the way that John suggested we do the `DateField` - that is, make the RCL export a wrapper component that just organizes the components, and then the implementer would use `TextField` components for the three inputs (would probably need variants). This would allow us to wrap each input in a redux-form `Field` component and would prevent us from having to split out the values in the RCL

patrickkim: Makes sense, I haven't figured out the cleanest way to allow the organizing of components.

src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
```diff
@@ -0,0 +1,22 @@
...
+  // beforeEach(() => {
+  //   defaultProps = {
+  //     // Add default props here
+  //   };
+  // });
+
+  it('renders', () => {
+    const wrapper = shallow(<CreditCardField />);
+
+    expect(wrapper.length).to.equal(1);
```
drewdrewthis: I feel like we can come up with a better standard test.

src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
drewdrewthis: @cisacke The whole test?

cisacke: @drewdrewthis just the commented out code - although I agree we should have a better standard test because the plop generated file isn't really doing anything at the moment

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
```diff
@@ -0,0 +1,48 @@
...
+    pointer-events: none;
+  }
+
+  &-line2 {
+    display: flex;
+  }
+
+  &-exp, &-sec { width: 50%; }
+
+  &-exp {
```
drewdrewthis: @patrickkim I know you love this set up with the `&-`. It's good, and it makes sense, but I honestly think it's unnecessarily verbose since we're using the css-modules. `.line2` won't conflict with any other classes since there's no global name space. 

The Upside to this pattern would be that, if we ever decided to throw out css-modules, we wouldn't need to convert any of our code. We should discuss this in standup.

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -226,7 +232,7 @@ $checkbox-border-color: color('neutral-5');
 
 // Checkbox List
 .checkboxList {
-  composes: formfield;
+  @include formfield;
```
drewdrewthis: @patrickkim Why are you not using a placeholder? j/w

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
```diff
@@ -0,0 +1,48 @@
...
+    border-right: 1px solid $form_border;
```
drewdrewthis: @patrickkim You might as well extract the whole `1px solid #color` into the variable.

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
patrickkim: old habits die hard.

src/molecules/formfields/shared/formfields.module.scss
patrickkim: it slipped my mind %formfield is the way to go.

drewdrewthis:
@patrickkim Hey man, where are you with this?

src/molecules/formfields/CreditCardField/index.js
patrickkim: Created a seperate ticket to handle this type of fieldgroup.

patrickkim:
Merging this in. Added a ticket to handle field groups properly which includes this and the date-field

PR #119, opened 2017-05-10
Add Footer Template component.

patrickkim:


src/atoms/Layout/index.js
```diff
@@ -81,7 +82,8 @@ function processChildren(props) {
       largeCols: largeCols[lgIdx],
       xLargeCols: xLargeCols[xlgIdx],
       xxLargeCols: xxLargeCols[xxlgIdx],
-      bottomSpacing
+      bottomSpacing,
+      fullwidth: fullWidthAll,
```
drewdrewthis: @patrickkim bhaha - - you had to go camelCase on the fullWidth. If you're going to do this, just change the other one so it's consistent. 

[fullwidth](https://en.wiktionary.org/wiki/fullwidth) is a word, though, and I think it looks much cleaner. Camel case is super ugly and harder to read.

src/atoms/Layout/index.js
```diff
@@ -170,6 +172,10 @@ Layout.propTypes = {
    */
   fullwidth: PropTypes.bool,
   /**
+   * removes left and right padding from Layout.
```
drewdrewthis: IMO this should be clearer -- "prevents `Layout` from passing left and right padding to all of its children (`<Cols />`). Will not remove padding that is explicitly added to a `<Col />`. << This should also be true if it's not.

src/templates/Footer/Readme.md
```diff
@@ -0,0 +1,10 @@
+Footer Example:
+
+```example
+    <div>
+      <Footer
+        phoneNumber='1(855) 695-2255'
```
drewdrewthis: Spacing between 1 and (

src/templates/Footer/__tests__/footer.spec.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+
+import Footer from 'templates/Footer';
+
+/* eslint-disable */
+jest.mock('atoms/Icon', () => {
```
drewdrewthis: This mock is going to be pretty common. Do we want to add it to the template, but commented out?

drewdrewthis: .. Probably not, but just a q.

src/templates/Footer/footer.module.scss
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  .help {
+    &-layout { justify-content: center; }
+    &-text { display: none; }
+  }
+
+  .trust {
+    &-logos { justify-content: center; }
+    &-links { justify-content: center; }
```
drewdrewthis: @patrickkim Consider combining these `justify-content` rules

src/templates/Footer/index.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import Button from 'atoms/Button';
+import Icon from 'atoms/Icon';
+import { Layout, Col } from 'atoms/Layout';
+import LinkWrapper from 'atoms/LinkWrapper';
```
drewdrewthis: I was thinking -- does it make sense to leverage Josh's context script for the npm publish to create a master index.js that we can just import these from? Instead of having 4 lines of imports, just have one?

src/templates/Footer/footer.module.scss
```diff
@@ -0,0 +1,115 @@
...
+    flex: 1 0 auto;
+    justify-content: space-between;
+  }
+
+  &-phone {
+    @include typography-6(false);
+  }
+
+  &-button {
+    $button_color: color('neutral-3');
```
drewdrewthis: @patrickkim Should be kebab case.

src/atoms/Layout/index.js
patrickkim: ok hard to argue with that changing.

src/templates/Footer/index.js
patrickkim: I'm in favor of less typing.

src/templates/Footer/__tests__/footer.spec.js
patrickkim: We should prevent the need for this mock.

src/templates/Footer/index.js
drewdrewthis: @patrickkim I think this might mess up our source map, though... I'm not sure.

src/templates/Footer/__tests__/footer.spec.js
drewdrewthis: So, the only way I can tell to do this AND maintain the automated behavior is with the script I spoke about that creates the loader inside of the `require`.. it's much less elegant, and this is actually what the mocks are specifically used for. We could extract it into a neat helper method, though, that will mock any components we tell it to.

PR #120, opened 2017-05-10
Fix tooltip and add modal state

drewdrewthis:
https://app.clubhouse.io/policygenius/story/1965/fix-tooltip-overflowing-page


src/atoms/Tooltip/index.js
```diff
@@ -2,22 +2,71 @@ import React from 'react';
...
+    this.setState({
+      modalIsOpen: false
+    });
+  }
+
+  render() {
+    const {
+      children,
+      className,
+      left,
```
patrickkim: This would be better a stateVariable `direction: left || right`. I'm guessing you don't want to be able to allow all left, center, right conditions?

PR #121, opened 2017-05-11
FlashMessages that animate!

patrickkim:


src/atoms/FlashMessage/flash_message.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  }
+  100% {
+    max-height: 200px;
+    opacity: 1;
+  }
+}
+
+.fadeIn { animation: fadeIn 600ms ease-in both; }
+.fadeInDown { animation: fadeInDown 600ms cubic-bezier(0.175, 0.885, 0.32, 1.275) both; }
+.slideDown { animation: slideDown 1000ms cubic-bezier(0.175, 0.885, 0.32, 1.275) both; }
```
drewdrewthis: @patrickkim We probably don't want to have to write this every time, and these should probably be standardized. I suggest importing keyframes into the global file and then having placeholders that can be used:

```
.fadeIn {
   @extend %fadeIn;
}
```

src/atoms/FlashMessage/index.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import styles from './flash_message.module.scss';
+
+function FlashMessage(props) {
+  return (
+    <div className={classnames(styles['flash-message'], styles[props.variant])}>
```
drewdrewthis: @patrickkim Why did you break away from our convention of deconstructing the `props` here? I know it's only two, but it's nicer if someone wants to add a third down the line.

src/molecules/asides/ImageAside/image-aside.module.scss
```diff
@@ -100,7 +100,9 @@ $aside-text-color:               color('neutral-2');
 }
 
 .small {
+  display: flex;
   align-items: center;
+  margin-left: 0;
```
drewdrewthis: Where was this margin coming from?

styleguide.config.js
```diff
@@ -46,14 +46,22 @@ module.exports = {
...
+          name: 'FlashMessage',
+          components: 'src/molecules/FlashMessage/**/*.js'
+        },
+        {
+          name: 'FormFields',
+          components: 'src/molecules/formfields/**/*.js'
+        },
+        {
+          name: 'Headers',
+          components: () => [ 'src/molecules/PriceDiscountHeader/index.js' ]
```
drewdrewthis: We shouldn't need to explicitly write these in every time. They should just get brought in by the "Other" category.

src/atoms/FlashMessage/Readme.md
```diff
@@ -0,0 +1,20 @@
...
+
+```example
+  <div>
+    <Button onClick={() => setState({ showMessage: !state.showMessage })}>Click me!</Button>
+    <Spacer xsmall />
+    { state.showMessage &&
+      <FlashMessage variant='fadeInDown'>
+        <ImageAside
+          small
+          basic={false}
```
drewdrewthis: @patrickkim We shouldn't have to do this. I made a ticket to refactor this: https://app.clubhouse.io/policygenius/story/2216/refactor-image-aside

src/atoms/FlashMessage/index.js
patrickkim: No reason, force of habit. Destructuring / es6 is still pretty new.

src/atoms/FlashMessage/flash_message.module.scss
patrickkim: Maybe when theres more than 1 component with animations that would be worth doing. Didn't want to over-engineer a global standard without a reference to other use cases.

styleguide.config.js
patrickkim: 💯

src/atoms/FlashMessage/Readme.md
patrickkim: 👍

src/molecules/asides/ImageAside/image-aside.module.scss
patrickkim: No idea... killed it.

PR #123, opened 2017-05-11
Prepare Credit Card field to work with redux form

cisacke:
@patrickkim got this working - would still like to pair tomorrow as it can definitely be better!


cisacke:
this is ready for 👀

src/molecules/formfields/CreditCardField/Readme.md
```diff
@@ -1,5 +1,27 @@
...
+          placeholder='Credit Card Number'
+          noBaseStyle
+        />
+      </CreditCardField.Top>
+
+      <CreditCardField.Bottom>
+        <TextField
+          placeholder='Expiration Date'
+          input={{
+            pattern: '[0-9]{13,16}'
```
jmcolella: what is this `pattern` property? I don't see it as a part of redux-form, just curious how we would use this in development

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
```diff
@@ -1,45 +0,0 @@
-$form_border: 1px solid color('neutral-5');
```
jmcolella: maybe we should keep this individual scss file instead of dumping everything in `shared`?

src/molecules/formfields/CreditCardField/Readme.md
cisacke: @jmcolella it's a html input attribute - you can pass a regex to control the input. @patrickkim added this in originally though & tbh I couldn't figure out if it was working correctly because it didn't seem to be limiting the input in any way

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
cisacke: yeah idk, I was just following the way we've been doing other formfields which is to keep all the specific styles in shared. it does seem like an anti pattern though, and that file is getting kinda big so we could extra all the specific styles into their own files & reference two sets of styles in each component?

jmcolella: I don't think we would need to reference two sets of styles. If we turn the shared classes into placeholders, then we can import that shared file into the individual scss files and use the placeholders as necessary

src/molecules/formfields/CreditCardField/Readme.md
jmcolella: hmm, is this necessary though? Do we need this plus validations, it seems redundant. also seems like someone could just delete this off the DOM and input whatever they like.

cisacke: probably a question for @patrickkim, but I agree that we probably don't need it with redux-form validations so I'll remove.

src/molecules/formfields/CreditCardField/credit_card_field.module.scss
cisacke: ok sounds like a good plan

src/molecules/formfields/CreditCardField/Readme.md
patrickkim: It's just a nice-to-have html5 input validation,  It gives you device/browser level validation messages that you don't have to make yourself.
 [https://www.w3schools.com/tags/att_input_pattern.asp](html5 input pattern)

patrickkim:
Merging.

PR #124, opened 2017-05-12
Fix button font weight

petrgazarov:
Changes in `src/molecules/lists/IconList/index.js` are generated by `yarn lint`


petrgazarov:
@drewdrewthis a quick 👓  when you have a chance

PR #128, opened 2017-05-12
Add Playback Card

cisacke:
@policygenius/front-end ready for 👀 
- adds playback card https://app.zeplin.io/project/58f0e0f5d78899b1b4dab89e/screen/590c8cb07fda054dae3509a0
- tweaks date field so it can be used with redux-form
- adds errors to `TextField` - this is a proof of concept more than anything, I think eventually we might be able to write a higher order component that takes care of errors so we don't have to replicate this logic everywhere, but this should be good enough for the ticket I'm working on

### DIFF
--- a/src/atoms/LinkWrapper/link_wrapper.module.scss
+++ b/src/atoms/LinkWrapper/link_wrapper.module.scss
@@ -1,6 +1,7 @@
 .link {
   color: color('neutral-4');
   display: inline-block;
+  text-decoration: underline;
 }
 
 @each $key, $value in $colors-map {
--- a/src/molecules/formfields/TextField/index.js
+++ b/src/molecules/formfields/TextField/index.js
@@ -12,26 +12,39 @@ function TextField( props ) {
     input,
     htmlFor,
     meta,
+    noBaseStyle
   } = props;
 
-  return (
-    <div className={classnames(styles.textfield, { [styles.focused]: meta && meta.active }, className)}>
-      {
-        label &&
-        <label
-          className={styles.label}
-          htmlFor={htmlFor}
-        >
-          { label }
-        </label>
-      }
+  const baseClassName = classnames(
+    { [styles.textfield]: !noBaseStyle },
+    { [styles.focused]: meta && meta.active && !noBaseStyle },
+    { [styles.hasError]: meta && meta.touched && meta.error },
+    className
+  );
 
-      <input
-        className={styles.input}
-        type='text'
-        placeholder={placeholder}
-        {...input}
-      />
+  return (
+    <div>
+      <div className={baseClassName}>
+        {
+          label &&
+          <label
+            className={styles.label}
+            htmlFor={htmlFor}
+          >
+            { label }
+          </label>
+        }
+
+        <input
+          className={styles.input}
+          type='text'
+          placeholder={placeholder}
+          {...input}
+        />
+      </div>
+      {meta && meta.touched && meta.error &&
+        <div className={styles.error}>{ meta.error }</div>
+      }
     </div>
   );
 }
@@ -67,6 +80,10 @@ TextField.propTypes = {
    * The props under the meta key are metadata about the state of this field that `redux-form` tracks.
    */
   meta: PropTypes.object,
+  /**
+   * Passing `noBaseStyle=true` will omit the base class style
+   */
+  noBaseStyle: PropTypes.bool
 };
 
 TextField.defaultProps = {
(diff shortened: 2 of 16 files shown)


src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -179,18 +181,20 @@ $checkbox-border-color: color('neutral-5');
...
+  &-input {
+    width: calc(100% / 3);
 
-    border-right: 1px solid color('neutral-4');
-  }
+    input {
+      width: 100%;
+    }
+
+    &:not( :last-child ) {
```
jmcolella: can we avoid using `:not` selector here? maybe make all of these styles part of the `datefield-input` and then the `:last-child` has specific styles? it's more verbose, but makes our styles more flexibile

src/organisms/cards/PlaybackCard/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+function PlaybackCard(props) {
+  const {
+    children,
+    header,
+    onClick
+  } = props;
+
+  const parts = separatePartials(children);
+  const renderPartial = partialRenderer(PlaybackCardPartials, styles.playbackCard);
```
jmcolella: i'm unclear why you are passing `styles.playbackCard` as a second arg? it seems `partialRenderer` only takes one

src/organisms/cards/PlaybackCard/playback_card.module.scss
```diff
@@ -0,0 +1,22 @@
+@import 'assets/stylesheets/global';
+
+$playback-card-border: 1px solid color('neutral-5');
+
+.playbackCard {
+  border-radius: 6px;
```
jmcolella: please use `ru(.25)`

src/organisms/cards/PlaybackCard/playback_card.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  border: $playback-card-border;
```
jmcolella: see variable comments above, I don't think this is necessary since it's used once

src/atoms/LinkWrapper/link_wrapper.module.scss
```diff
@@ -1,6 +1,7 @@
 .link {
   color: color('neutral-4');
   display: inline-block;
+  text-decoration: underline;
```
jmcolella: @drewdrewthis should we do this?

src/molecules/formfields/TextField/index.js
```diff
@@ -12,26 +12,39 @@ function TextField( props ) {
     input,
     htmlFor,
     meta,
+    noStyle
   } = props;
 
+  const baseClassName = classnames(
```
jmcolella: I feel like we should do this as a standard, much cleaner

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -13,6 +13,8 @@ $checkbox-disabled-color: color('neutral-3');
 $checkbox-disabled-sign-color: color('neutral-3');
 $checkbox-background-color: color('neutral-8');
 $checkbox-border-color: color('neutral-5');
+$field-error-color: color('accent-2');
```
jmcolella: not sure these variables are necessary since you're just using them once

src/atoms/LinkWrapper/link_wrapper.module.scss
cisacke: @jmcolella the `text-decoration` for the anchor tags is currently coming from the agent user stylesheet, which only adds it if there's a `href` attribute on the tag. I think we are going to end having a lot of LinkWrappers that only take `onClick` which is why I added it, although there might be a better/different solution?

src/organisms/cards/PlaybackCard/index.js
cisacke: my bad, I was trying something out & forgot to remove this

src/molecules/formfields/shared/formfields.module.scss
cisacke: 👍 good point, i'll take them out

src/molecules/formfields/TextField/index.js
```diff
@@ -67,6 +80,10 @@ TextField.propTypes = {
    * The props under the meta key are metadata about the state of this field that `redux-form` tracks.
    */
   meta: PropTypes.object,
+  /**
+   * Passing `noStyle=true` will omit the base class style
+   */
+  noStyle: PropTypes.bool
```
cisacke: yeah that's definitely a better name, i'll change to `noBaseStyle`

src/molecules/formfields/shared/formfields.module.scss
cisacke: 👍

cisacke:
@jmcolella thanks for your feedback on this - gonna merge so I can keep going with my renters ticket but lmk if there's anything else I can do here!

PR #129, opened 2017-05-14
Add links to footer, Add Navigator.footer, Add Footer to Navigator.

patrickkim:
…ample

### DIFF
--- a/src/templates/Footer/index.js
+++ b/src/templates/Footer/index.js
@@ -9,12 +9,33 @@ import LinkWrapper from 'atoms/LinkWrapper';
 
 import styles from './footer.module.scss';
 
+function renderLinks(linksArr) {
+  if (!linksArr || linksArr.length === 0) return null;
+
+  function renderLink(link) {
+    return (
+      <li className={styles['trust-link']}>
+        <LinkWrapper href={link.href}>{link.text}</LinkWrapper>
+      </li>
+    );
+  }
+
+  return (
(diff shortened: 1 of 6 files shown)


src/templates/Footer/index.js
```diff
@@ -9,13 +9,34 @@ import LinkWrapper from 'atoms/LinkWrapper';
 
 import styles from './footer.module.scss';
 
+function renderLinks(linksArr) {
+  if (!linksArr || linksArr.length === 0) return null;
+
+  function renderLink(link) {
+    return (
+      <li className={styles['trust-link']}>
+        <LinkWrapper href={link.href || '#'}>{link.text}</LinkWrapper>
```
jmcolella: can we validate in prop types that there is always an `href`? seems counter intuitive to have footer links with dummy `href`s

src/templates/Footer/index.js
```diff
@@ -80,10 +99,18 @@ Footer.propTypes = {
    * Function to trigger chat.
    */
   onClickChat: PropTypes.func,
+  /**
+   * Array of Links `[ { text: 'link text', href: 'http://link.com' } ]`
+   */
+  links: PropTypes.array,
```
jmcolella: could do `PropTypes.arrayOf(PropTypes.shape({ text: PropTypes.string, href: PropTypes.string.isRequired }))`. it's verbose, but could be helpful for ensuring an `href` is passed for every link

PR #135, opened 2017-05-15
Update footer

petrgazarov:
- Fixes minor styling and markup
- updates props

### DIFF
--- a/src/templates/Footer/footer.module.scss
+++ b/src/templates/Footer/footer.module.scss
@@ -1,17 +1,22 @@
 .footer {
   color: color('neutral-3');
-  background-color: color('neutral-7');
+  background-color: color('neutral-6');
   border-top: 1px solid color('neutral-5');
 
-  padding: ru(1) ru(.75);
+  padding: ru(1) ru(2);
 
   &-layout {
     display: flex;
     flex-direction: row-reverse;
+    max-width: '1000px';
   }
 
   &-trust {
     flex: 1 0 auto;
+  }
+
+  &-help {
+    max-width: '330px';
   }
 }
 
@@ -26,6 +31,8 @@
 
   &-phone {
     @include typography-6(false);
+    @include typography-semibold;
+    color: inherit;
   }
 
   &-button {
@@ -102,7 +109,9 @@
 
   .help {
     &-layout { justify-content: center; }
-    &-text { display: none; }
+    &-text {
+      margin-bottom: ru(1);
+    }
   }
 
   .trust {
--- a/src/templates/Footer/index.js
+++ b/src/templates/Footer/index.js
@@ -31,24 +31,25 @@ function Footer(props) {
   const {
     className,
     phoneNumber,
-    text,
+    emailAddress,
     links,
     onClickChat,
   } = props;
-
 
   return (
     <div className={classnames(styles['footer'], className)}>
       <Layout
         className={styles['footer-layout']}
-        mediumCols={[ 5 ]}
+        mediumCols={[ 6 ]}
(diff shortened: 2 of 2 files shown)


src/templates/Footer/index.js
```diff
@@ -109,13 +116,13 @@ Footer.propTypes = {
     })
   ),
   /**
-   * Small Footer text.
+   * set the mailto: email address
```
drewdrewthis: @petrgazarov Sorry -- this should read "Sets the `mailto:` email address"

src/templates/Footer/index.js
```diff
@@ -109,13 +116,13 @@ Footer.propTypes = {
...
    */
-  text: PropTypes.string,
+  emailLink: PropTypes.string,
```
drewdrewthis: @petrgazarov This should probably be something like `emailAddress` so that people do think that they should include the whole `mailto:team....`

src/templates/Footer/footer.module.scss
```diff
@@ -26,6 +26,8 @@
 
   &-phone {
     @include typography-6(false);
+    color: inherit;
+    font-weight: 700;
```
drewdrewthis: @petrgazarov This font-weight should be set with our mixin. We don't explicitly write font-weights anymore.

src/templates/Footer/index.js
```diff
@@ -41,14 +41,18 @@ function Footer(props) {
...
-        mediumCols={[ 5 ]}
+        mediumCols={[ 6 ]}
         smallCols={[ 12 ]}
         fullwidthAll
+        style={{ maxWidth: '1000px' }}
       >
-        <Col className={classnames(styles['footer-help'], styles['help'])}>
+        <Col
+          className={classnames(styles['footer-help'], styles['help'])}
+          style={{ maxWidth: '330px' }}
```
drewdrewthis: @petrgazarov Not sure why you set this, but since this is actually in the RCL / Athenaeum you should do this in the scss. We only exposed this prop so that users can update it in their projects.

petrgazarov:
@drewdrewthis Thanks for the comments, I think this is good to be re-reviewed.

src/templates/Footer/index.js
```diff
@@ -31,24 +31,25 @@ function Footer(props) {
   const {
     className,
     phoneNumber,
-    text,
+    emailAddress,
```
patrickkim: Are we sure we want to require exposing an email address with `mailto:` over having a generic onclick? 

- explicit exposed email addresses on a site thats search engine crawl-able will always render that mailbox useless with the sheer amount of spam.
- allows sending email via a third-party like intercom

If we allow one or the other or both the `onClickChat` prop should either be removed or made to work.

petrgazarov: > explicit exposed email addresses on a site thats search engine crawl-able will always render that mailbox useless with the sheer amount of spam.

Has this been an issue? I don't mind changing it, but I'm curious why. We have our email exposed in multiple places in the app (including in the existing Footer).

> If we allow one or the other or both the onClickChat prop should either be removed or made to work.

Can you please clarify? `onClickChat` works now if passed. If it isn't passed, the button doesn't render.

patrickkim: Sorry Petr, `onClickChat` I was mistaken ignore what I said about that.

`mailto` spam has pretty common for many years. There are tons of email obfuscation scripts out there because of it. If we haven't had a problem that's good. It must mean our spam filter is very good, it's not being crawled or it's not big enough a problem for people to complain about.

PR #136, opened 2017-05-15
Fix checkout layout

drewdrewthis:
- Add footer to checkout
- Fix styles/layout for checkout
- Fix fullscreen view on Styleguidist
- Update footer background color


patrickkim:
I think this is the `<Icon />` component failing the spec. I just checked-in a fix that shouldn't require this anymore.

src/templates/CheckOut/CheckOutPartials.js
```diff
@@ -80,10 +74,16 @@ export default {
 
     return (
       <Col className={styles['main-sidebar']}>
-        <div className='sidebar'>{children}</div>
```
patrickkim: remove line 63 `expect(wrapper.find('.sidebar')).to.have.length(1);` in checkout.spec.js

PR #138, opened 2017-05-16
Update images manifest loader to make Icon component more robust.

patrickkim:
Prevents having to mock `<Icon />` in each spec.


src/atoms/Icon/index.js
```diff
@@ -18,9 +18,7 @@ function Icon( props ) {
       className={classnames(styles['icon-wrapper'], { [styles.clickable]: onClick }, className)}
       onClick={onClick}
     >
-      <SVGInline
-        svg={icon}
-      />
+      { icon && <SVGInline svg={icon} /> }
```
patrickkim: Short circuit the component if there is no icon to pass in.

src/assets/images/index.js
```diff
@@ -1,9 +1,23 @@
...
+    return Object.assign({}, res, fileObj);
+  };
+
+  return context.keys().reduce(fileManifest, {});
+}
+
+let svgContext;
+
+// Skip context loading in tests. require.context is not available in jest.
+if (module.hot) {
```
patrickkim: Seems a little hacky but this was cleaner than mocking child dependencies in the specs.

drewdrewthis: @patrickkim Why did you rewrite the `fromPairs` code into a much more verbose / complex function? If you're just going to test for the method, you could have simply wrapped everything in the conditional.

src/atoms/Icon/index.js
drewdrewthis: :+1:

styleguide.config.js
```diff
@@ -14,91 +14,41 @@ module.exports = {
...
-        {
-          name: 'TextLockups',
-          components: 'src/molecules/TextLockup/**/*.js'
-        },
-        {
-          name: 'Other',
-          components: () => [ 'src/molecules/StepProgress/index.js' ]
-        }
-      ]
+      components: 'src/molecules/**/index.js',
```
drewdrewthis: This removes the sectioning on things like `FormFields` and `TextLockups`. These were there on purpose.

drewdrewthis:
👀

src/assets/images/index.js
patrickkim: :finnadie:

PR #144, opened 2017-05-16
Update Add Card

jmcolella:
@drewdrewthis Fixes issue we debugged


src/organisms/forms/shared/forms.module.scss
```diff
@@ -23,6 +23,8 @@ $icon-size:        ru(1);
   border-bottom-width: $border-width;
   padding-top: ru(1);
   padding-bottom: ru(1.75);
+  padding-left: ru(1);
```
patrickkim: Shouldn't this be done with `<Layout/>`

PR #153, opened 2017-05-18
Add input mask to textfield for phone numbers

cisacke:


cisacke:
@patrickkim @drewdrewthis adding an input mask for phone numbers, should be a quick 👀  when you have a second. Thanks!

patrickkim:
👀

src/molecules/formfields/TextField/index.js
```diff
@@ -35,13 +37,23 @@ function TextField( props ) {
...
           </label>
         }
-
-        <input
-          className={styles.input}
-          type='text'
-          placeholder={placeholder}
-          {...input}
-        />
+        { phone ?
```
patrickkim: Mask seems useful, I think this should be switch out on mask and have the mask pattern and char as an available prop on TextField

cisacke:
@patrickkim made those changes

PR #154, opened 2017-05-18
Add color prop to TextComponent

patrickkim:


src/atoms/Color/text_color.module.scss
```diff
@@ -0,0 +1,3 @@
+@each $key, $value in $colors-map {
```
jmcolella: why did you add this file? just curious

drewdrewthis:
@patrickkim Before you merge

src/atoms/Color/text_color.module.scss
patrickkim: my mistake. got in there by accident.

patrickkim:
I was deferring that decision for now. I added a chore to handle this in a cleaner way on the css side.

PR #155, opened 2017-05-18
Update for better react-styleguidist use

drewdrewthis:
#Blocked by react-styleguidist update:

https://github.com/policygenius/react-styleguidist/pull/5


styleguide.config.js
```diff
@@ -31,75 +31,84 @@ module.exports = {
...
-          components: 'src/organisms/tables/**/index.js'
-        }
-      ]
-    },
-    {
-      name: 'Templates',
-      components: 'src/templates/**/index.js',
-    },
+      name: 'Design',
+      components: 'src/atoms/**/index.js'
```
jmcolella: is this just a placeholder?

drewdrewthis: @jmcolella Yes. We don't know how they'll want to organize everything, so I just left it. It's also looking like you can't include components twice, so we'll have to figure something out.

webpack.config.base.js
```diff
@@ -18,20 +23,12 @@ module.exports = options => ({
...
   },
   module: {
     rules: get(options, 'module.rules', []).concat([
       {
         test: /\.js$/,
-        exclude: [
-          path.resolve(__dirname, 'node_modules'),
-          /__tests__/
-        ],
+        include: includePaths,
```
jmcolella: I might change this to `options.includePaths` and just keep the `includePaths` constant in the `dev` config since I'm not sure we need to include `styleguide_assets` for prod.

drewdrewthis: hmm. Interesting. Sounds good.

patrickkim:
Lets get this going.

PR #157, opened 2017-05-18
RadioCard is sectionable

patrickkim:


src/organisms/cards/RadioCard/radio_card.module.scss
```diff
@@ -71,6 +73,42 @@
...
+
+  &-item {
+    flex: 1 1;
+    margin: 0;
+    padding: ru(.75);
+  }
+
+  &-value {
+    @include typography-4;
+    font-weight: 600;
```
drewdrewthis: @patrickkim Please use the mixin for font weight.

src/organisms/cards/RadioCard/radio_card.module.scss
```diff
@@ -71,6 +73,42 @@
...
+    position: relative;
+    padding: ru(2.75) ru(1) ru(1);
+  }
+}
+
+.sections {
+  @include typography-7;
+
+  display: flex;
+  margin: auto ru(-1) ru(-1);
```
drewdrewthis: Do we really need to use negative margins everywhere for this? I think it would be preferable to just remove the padding, instead of canceling it out with margins.

If we have to use negative margins, it would be better to make them use variables that are tied to the padding they are negating so that they will always be in sync.

src/organisms/cards/RadioCard/radio_card.module.scss
```diff
@@ -126,4 +164,28 @@
...
+      margin: ru(.5);
+    }
+
+    .title + .description {
+      margin-top: ru(-.5);
+    }
+
+    .content {
+      flex: 1 0;
+      margin: ru(-.5) ru(-.5) ru(-.5) ru(.5);
```
drewdrewthis: See other comments about negative margins.

src/organisms/cards/RadioCard/index.js
```diff
@@ -35,6 +59,7 @@ function RadioCard(props) {
           <h4 className={styles['title']}>{label}</h4>
           { description && <p className={styles['description']}>{description}</p> }
```
drewdrewthis: Dot notation: `styles.description`

drewdrewthis:
Also, the icons aren't showing up here. It seems like you need to rebase?

http://rcl.policygenius.com/radio-sections-cfafb5ae-94a2-45c1-9aad-b7ccf0a3e043/index.html

patrickkim:
I've confirmed that all the Icons show up after a rebase.Good call on all the comments minus the dot notation. I copied & pasted them in from the old styleguide. Accept nothing, question everything. 
:100: :feelsgood:

PR #158, opened 2017-05-18
Add auto sizing on icons, Add example on PlaybackCardWrapper.

patrickkim:


src/atoms/Icon/index.js
```diff
@@ -33,6 +40,10 @@ Icon.propTypes = {
    */
   icon: PropTypes.string,
   /**
+   * The icon you'd like to display.
```
cisacke: change this for `sizeAuto`!

src/organisms/cards/PlaybackCardWrapper/index.js
```diff
@@ -11,16 +13,33 @@ function PlaybackCardWrapper(props) {
...
+    unit,
     footer,
     footerOnClick,
   } = props;
 
+  const renderHeader = () => {
+    if (!logo || !amount) return null;
+
+    return (
+      <Layout
```
cisacke: it looks like the padding is a little off on the playback card wrapper - should be ru(1.5), and then there should be a little more padding under the logo

PR #163, opened 2017-05-19
TextComponent is now Text. Add shared colors css obj.

patrickkim:


src/atoms/Text/index.js
```diff
@@ -0,0 +1,82 @@
...
+   * provided in the component's index.js file.
+   */
+  className: PropTypes.string,
+  /**
+   * You can use any html element text tag
+   */
+  tag: PropTypes.string,
+
+  /**
+   * determines typography class
```
drewdrewthis: Capital 'D'?

src/atoms/Text/index.js
```diff
@@ -0,0 +1,82 @@
...
+   */
+  type: PropTypes.number,
+
+  /**
+   * Font weight: light|regular|bold
+   * - light = 300
+   * - semibold = 500
+   * - bold = 600
```
drewdrewthis: This is really confusing. My understanding is that we only have light, regular, and semibold options. `regular` should probably be the default.

src/atoms/TextComponent/Readme.md
```diff
@@ -1,14 +1 @@
...
-    <TextComponent type={1}>Type 1</TextComponent>
-    <TextComponent type={2}>Type 2</TextComponent>
-    <TextComponent type={3}>Type 3</TextComponent>
-    <TextComponent type={4}>Type 4</TextComponent>
-    <TextComponent type={5}>Type 5</TextComponent>
-    <TextComponent type={6}>Type 6</TextComponent>
-    <TextComponent type={7}>Type 7</TextComponent>
-  </div>
-```
+To be deprecated. Use `<Text />`
```
drewdrewthis: :+1:

src/atoms/Text/__tests__/text.spec.js
```diff
@@ -0,0 +1,26 @@
...
+import { shallow, mount } from 'enzyme';
+
+import Text from 'atoms/Text';
+
+describe('<Text />', () => {
+
+  it('renders', () => {
+    const wrapper = shallow(<Text />);
+
+    expect(wrapper.length).to.equal(1);
```
drewdrewthis: Please change this to the `expect(wrapper.type()).to.equal(Text);` that we'd discussed.

src/atoms/TextComponent/index.js
```diff
@@ -1,92 +1,4 @@
...
-};
-
-
-TextComponent.defaultProps = {
-  tag: 'p',
-  type: 6
-};
-
-export default TextComponent;
+export default props => <Text {...props} />;
```
drewdrewthis: Slick!

drewdrewthis: Can you not just `export default Text`?

src/organisms/cards/RadioCard/__tests__/radio_card.spec.js
```diff
@@ -7,7 +7,7 @@ import RadioCard from 'organisms/cards/RadioCard';
 describe('<RadioCard />', () => {
 
   it('renders', () => {
-    const wrapper = shallow(<RadioCard label='its a label' input={{}} />);
+    const wrapper = shallow(<RadioCard label='its a label' input={{}} radioValue='something' />);
 
     expect(wrapper.length).to.equal(1);
```
drewdrewthis: Same comment as above.

src/atoms/Text/index.js
patrickkim: It's confusing because the doc was wrong.

src/atoms/Text/__tests__/text.spec.js
patrickkim: `type` doesn't work like that. In most cases type will be `<div>`, the wrapping element.  If you had a component that returns `<Something>thing</Something>` its type will be `Something`.

src/atoms/TextComponent/index.js
patrickkim: Either the linter was failing or it didn't work, I forget which didn't give it much thought.

PR #166, opened 2017-05-20
WIP Silence ru warnings

patrickkim:
I bumped up the units and changed them when it made sense. We need better docs on this standard. When and where is this used? Sometimes converting to `ru` doesn't even make sense.

### DIFF
--- a/src/atoms/Button/buttons.module.scss
+++ b/src/atoms/Button/buttons.module.scss
@@ -114,7 +114,7 @@ $button-solid-background-tertiary-color:   #f08f5c;
   color: color('neutral-4');
   border: 1px solid color('neutral-5');
   box-shadow: 0 4px 0 color('neutral-6');
-  padding-bottom: ru(.375);
+  // padding-bottom: ru(.375);
 
   &:focus {
     outline: none;
@@ -153,7 +153,7 @@ $button-solid-background-tertiary-color:   #f08f5c;
   @include hover-action(0, rem-calc(5px), $button-action-disabled-border-color,
   $button-action-disabled-background-color, $button-action-disabled-color);
   background-color: $button-action-disabled-background-color;
-  box-shadow: 0 ru(.208) 0 $button-action-disabled-border-color;
+  box-shadow: 0 5px 0 $button-action-disabled-border-color;
   color: $button-action-disabled-color;
   transition: none;
   cursor: not-allowed;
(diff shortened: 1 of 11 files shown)


src/atoms/Button/buttons.module.scss
```diff
@@ -114,7 +114,7 @@ $button-solid-background-tertiary-color:   #f08f5c;
   color: color('neutral-4');
   border: 1px solid color('neutral-5');
   box-shadow: 0 4px 0 color('neutral-6');
-  padding-bottom: ru(.375);
+  // padding-bottom: ru(.375);
```
drewdrewthis: If this isn't necessary, we should remove the comment.

src/atoms/Button/buttons.module.scss
```diff
@@ -153,7 +153,7 @@ $button-solid-background-tertiary-color:   #f08f5c;
   @include hover-action(0, rem-calc(5px), $button-action-disabled-border-color,
   $button-action-disabled-background-color, $button-action-disabled-color);
   background-color: $button-action-disabled-background-color;
-  box-shadow: 0 ru(.208) 0 $button-action-disabled-border-color;
+  box-shadow: 0 5px 0 $button-action-disabled-border-color;
```
drewdrewthis: We should still use `rem-calc`.

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -407,7 +407,7 @@ $focused-border-color: color('brand-3');
     }
 
     span {
-      margin-top: ru(.125);
+      margin-top: 0;
```
drewdrewthis: Does the `span` come with a margin by default? Can we just remove this styling altogether?

src/organisms/cards/RadioCard/radio_card.module.scss
```diff
@@ -108,7 +108,7 @@ $border-style: 1px solid color('neutral-5');
     @include typography-6;
     @include typography-semibold;
 
-    margin: 0 0 ru(.125);
+    margin: 0;
```
drewdrewthis: We should probably use `typography-6(false)` instead here.

patrickkim:
the spacing is definitely off. There a lot of going in and pixel finessing that will have to happen to get them all silenced. Not high value so leaving as WIP for now. will revisit later.

PR #174, opened 2017-05-22
Text Component got some upgrades

patrickkim:


src/atoms/Text/index.js
```diff
@@ -5,19 +5,77 @@ import classnames from 'classnames';
...
+    variant: 'strikethrough'
+  });
+
+  return Text(newProps);
+}
+
+function Sup(props) {
+  const newProps = Object.assign({}, props, {
+    tag: 'sup',
+    variant: 'sup'
```
drewdrewthis: @patrickkim This will create duplicate class, I believe, since each Text component gets a styles[tag] and a styles[variant]

src/atoms/Text/index.js
```diff
@@ -5,19 +5,77 @@ import classnames from 'classnames';
...
+  const newProps = Object.assign({}, props, {
+    tag: 'em',
+  });
+
+  return Text(newProps);
+}
+
+function Strong(props) {
+  const newProps = Object.assign({}, props, {
+    tag: 'strong',
```
drewdrewthis: This will automatically bump the weight, I believe, and the `strong` weight isn't set in the CSS. Perhaps add it to `pg-normalize`?

src/atoms/Text/index.js
patrickkim: true good catch.

src/atoms/Text/index.js
```diff
@@ -5,19 +5,85 @@ import classnames from 'classnames';
...
+
+  if (semibold) return 'strong';
+  if (variant === 'italics') return 'em';
+  if (variant === 'strikethrough') return 'span';
+
+  return tag;
+};
+
+const getWeight = ({ semibold, light }) => {
+  if (semibold) return 'semibold';
```
drewdrewthis: Nice!

src/atoms/Text/index.js
```diff
@@ -5,19 +5,85 @@ import classnames from 'classnames';
...
+const getTag = (props) => {
+  const {
+    tag,
+    semibold,
+    variant,
+  } = props;
+
+  if (semibold) return 'strong';
+  if (variant === 'italics') return 'em';
+  if (variant === 'strikethrough') return 'span';
```
drewdrewthis: Should we make the `italics` and `strikethrough` bools as well, since we have these neat "get" functions to figure things out?

src/atoms/Text/text.module.scss
```diff
@@ -4,16 +4,17 @@
...
+  @include typography-regular;
 }
 
 .light {
   @include typography-light;
 }
 
-.regular {
-  @include typography-regular;
+.strong,
```
drewdrewthis: I meant in the base styles. A `strong` tag should never have a different weight than a `.strong`

PR #177, opened 2017-05-22
Fix Bucket-spacing left padding & on small screens.

patrickkim:


src/atoms/Layout/layout.module.scss
```diff
@@ -45,11 +55,16 @@ $border: 1px solid color('neutral-6');
     [class*='col-'] {
       border-left: $border;
       border-left: $border;
-      padding: ru(.75) ru(.5);
+      padding: 0 ru(.5) ru(1.25);
```
cisacke: @patrickkim we still want the padding around the children that are in the layout - just the left padding on the first child should be 0, and then on small screen sizes there should be ru(1) margin/padding bottom - looks like at the moment it's ru(0.5)

cisacke: also this is what I'm seeing on my renters branch - were you able to replicate in the RCL?
screen shot 2017-05-22 at 4 54 10 pm

PR #185, opened 2017-05-24
Add warnings

jmcolella:
@drewdrewthis @patrickkim @cisacke Adds warnings to certain fields

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "jest": "^19.0.2",
     "lodash": "^4.17.4",
     "react-input-mask": "^0.8.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -92,6 +91,7 @@
     "react-dom": "^15.5.4",
     "react-hot-loader": "^1.3.1",
     "react-modal": "^1.7.7",
+    "react-stickynode": "^1.3.1",
```
patrickkim: Is this meant to be here?

patrickkim: otherwise looks good.

PR #191, opened 2017-05-25
Stylized Dom Element components

patrickkim:


src/assets/stylesheets/base.scss
```diff
@@ -1,2 +1,3 @@
-@import 'assets/stylesheets/base_styles/sanitize';
+// @import 'assets/stylesheets/base_styles/sanitize';
```
patrickkim: There seems to some styling that comes with sanitize, I propose we just start everything from scratch with reset. thoughts?

drewdrewthis: @patrickkim I did a bit of research on this, and the best reasoning to use sanitize is presented on their website: https://jonathantneal.github.io/sanitize.css/

If you still think we should change over to reset.css, I'm down to discuss.

src/atoms/Element/Readme.md
```diff
@@ -0,0 +1,94 @@
...
+    <Element tag='u'>spacer</Element> set above me</Element>
+  </div>
+```
+
+```js
+  // If you use the Elemental module, you can write familiar semantic Html.
+  // The above example would be written as:
+  <div>
+    <H1 color='accent-3' typography={3}>h1 tags here.</H1>
+    <P color='neutral-4' spacer={12} typography={7}>custom p tag.</P>
```
drewdrewthis: @patrickkim Do you think you could explain `spacer` a bit better? Especially the fact that it is only bottom spacing?

src/atoms/Element/Readme.md
```diff
@@ -0,0 +1,94 @@
...
+      Table: (props) => <Table {...props} />
+    }
+  */
+```
+
+`styled(string)`
+
+Returns an `<Element />` with the passed in string argument as its tag. You can pass in an HTML tag name or custom name.
+```html
+  styled(tagName) = <Element tag='tagName' />
```
drewdrewthis: @patrickkim I don't understand why you created this method? Or why / how it would be used without passing in `children`?

src/atoms/Element/index.js
```diff
@@ -0,0 +1,100 @@
...
+   * 7: 13px
+   */
+  typography: PropTypes.oneOf([ 0, 1, 2, 3, 4, 5, 6, 7 ]),
+  /**
+   * Controls spacing below this element. Default is at 0. Increases by ru(.25) until spacer={10}.
+   * spacer={11} yields ru(3). /n
+   * spacer={12} yeilds ru(6).
+   */
+  spacer: PropTypes.oneOf([ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 ]),
+  color: PropTypes.oneOf([
```
drewdrewthis: I tried to get this to be created dynamically based on class options --- it works for validation, but doesn't display correctly in the documentation. Shame :/

i.e. `color: PropTypes.oneOf(colors.keys())` <-- works in validation, not in documentation

src/atoms/Spacer/index.js
```diff
@@ -14,20 +14,16 @@ function Spacer( props ) {
...
-    { [styles.medium]: medium },
-    { [styles.large]: large },
-    { [styles.xLarge]: xLarge },
-  ).split( ' ' );
+  const classes = [
+    xSmall && styles['x-sm'],
+    small && styles['sm'],
+    medium && styles['med'],
+    large && styles['lg'],
+    xLarge && styles['x-lg'],
```
drewdrewthis: @patrickkim If you're going to change this syntax here, I think it should also be changed in the `Layout` component for consistency.

src/atoms/Element/Readme.md
patrickkim: My documentation here makes it confusing. Would this make it clearer?

`const AnewComponentName = styled('AnewComponentName'); // returns <Element tag='AnewComponentName />`

src/atoms/Spacer/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 
-import styles from './spacer.module.scss';
+import styles from './spacer.scss';
```
drewdrewthis: @patrickkim This should maybe be called `spacings.scss` for convention? And potentially put in a branding directory with colors/typography. Thoughts?

drewdrewthis: Also, I suggest importing it as `spacing` for consistency with the other components

src/atoms/Spacer/spacer.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.top-11 { margin-top: ru(3); }
+.bot-11 { margin-bottom: ru(3); }
+
+.top-12 { margin-top: ru(6); }
+.bot-12 { margin-bottom: ru(6); }
+
+/* -- named spacer sizes --
+  Named spacer sizes uses @extend to save some
+  compiled space and provide you dev a map to the unit
```
drewdrewthis: "provide you dev a map to the unit"? I don't understand this sentence.

src/atoms/Element/Readme.md
patrickkim: I wrote a little more into detail about this on the prop description of spacer.

src/atoms/Spacer/spacer.scss
drewdrewthis: I kind of get why you're giving the dev the option to use "any" spacing option.. but it seems like giving them explicit permission to not use one of the 5. If you want to do this, perhaps consider giving the dev the option to simply use a number with the `space` prop in addition to the string `x-sm`, and then just inlining the conversion to `ru`? Also, it's not clear when / how these classes are used.

src/molecules/ComparisonRowItem/Readme.md
```diff
@@ -0,0 +1,15 @@
+### ComparisonRowItem
+
+This is a responsive helper component used with `<ComparisonTable />` to render cells that display differently on smaller devices. On larger screen it renders as  a checkmark or x icon. On small screens it renders as a `<LogoText />`.
+
+```example
+    const IconPg = <Icon icon='pgLogoOrange' width='120px' height='24px' />;
+    const IconMetLife = <Icon icon='metlife' width='120px' height='24px' />;
+    const LogoPg = <LogoTextA line1='header' logo={IconPg} />;
+    const LogoMetLife = <LogoTextA line1='header' logo={IconMetLife} />;
```
drewdrewthis: I don't entirely understand this.

src/molecules/ComparisonRowItem/index.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+ComparisonRowItem.propTypes = {
+  /**
+   * This prop will add a new className to any inherent classNames
+   * provided in the component's index.js file.
+   */
+  className: PropTypes.string,
+  /**
```
drewdrewthis: We were talking about standardizing this -- about half of our files put a space between propTypes. Thoughts?

src/atoms/Spacer/index.js
patrickkim: There's a branding directory?

src/molecules/LockUps/TextD/index.js
```diff
@@ -0,0 +1,26 @@
...
+  spacer: PropTypes.number,
+  line1: PropTypes.node,
+  line2: PropTypes.node,
+};
+
+TextD.defaultProps = {
+  spacer: 2,
+};
+
+export default TextD;
```
drewdrewthis: It would be nice if we had a way of handling TextLockups similarly to the way we handle icons. Doing it this way could easily get unweildy and make it difficult to update / pivot our TextLockups in the future.

src/organisms/cards/PercentageCard/index.js
```diff
@@ -31,18 +32,12 @@ function PercentageCard( props ) {
 
       <div className={styles['row']}>
         <div className={styles['col']}>
-          <TextLockup
-            variant='A'
-            header={header1}
-            subheader={subheader1}
-          />
+          <H3 typography={4} weight='light'>{header1}</H3>
```
drewdrewthis: @patrickkim One of the reasons we didn't explicitly use specific tags in the components is because this should really be an implementation choice depending on the other headers on the page and SEO needs. The user should be able to use the percentage card with any tags

src/organisms/cards/PlaybackCardWrapper/Readme.md
```diff
@@ -49,31 +45,27 @@ PlaybackCardWrapper Example:
...
           </div>
         }
 
         rightPanel={
           <div>
-            <TextLockup
-              variant='G'
-              header='Date of Birth'
-              subheader='9/19/1987'
+            <TextD
```
drewdrewthis: @patrickkim Yeah -- do you feel strongly about this explicitly named TextLockup components convention?

src/atoms/Spacer/index.js
drewdrewthis: Not currently, but there probably should be, with the idea that that dir can just be swapped out to get a different branding across the board.

drewdrewthis:
@patrickkim Also, could you please change `    "policygenius-react-styleguidist": "^0.0.7",
` => `    "policygenius-react-styleguidist": "latest"` in this PR?

src/atoms/Spacer/spacer.scss
patrickkim: Fair point.

src/molecules/ComparisonRowItem/Readme.md
patrickkim: This was little rewrite pass the linter. It makes more sense as a multi-line JSX but the linter wont allow it. Basically, LogoText takes in a logo node.

src/atoms/Element/index.js
patrickkim: Yea would be nice if this were possible

src/molecules/ComparisonRowItem/index.js
patrickkim: Nice to do but very little gain? maybe an intern can do this?

PR #195, opened 2017-05-26
Update error messages

jmcolella:
also Adds Transition component

@patrickkim @cisacke @drewdrewthis please CR

I know this solution isn't amazing, but I think it works for now and we can fine tune later

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "react-modal": "^1.7.7",
     "react-stickynode": "^1.3.1",
     "react-test-renderer": "^15.5.4",
+    "react-transition-group": "^1.1.3",
     "remark-cli": "^3.0.0",
     "remark-lint": "^6.0.0",
     "remark-lint-code": "^2.0.0",
--- a/src/atoms/ErrorMessage/Readme.md
+++ b/src/atoms/ErrorMessage/Readme.md
@@ -1,0 +1,38 @@
+ErrorMessage example:
+
+Note: You do not need to specifiy `ErrorMessage.default` in implementation. This is only needed for use in this guide.
+Simply use `<ErrorMessage {...props} />`
+
+```example
+  const ErrorMessage = require('atoms/ErrorMessage');
+
+  class ErrorMessageExample extends React.Component {
+    constructor() {
+      super()
+
+      this.state = { visible: false }
+
+      this.onClick = this.onClick.bind(this);
+    }
+
+    onClick() {
+      this.setState({ visible: !this.state.visible });
+    }
+
+    render() {
+      return (
+        <div>
+          <Button
+            onClick={this.onClick}
+          >
+            Click Me to reveal error message
+          </Button>
+
+          <ErrorMessage.default condition={this.state.visible} message={'Error'} />
+        </div>
+      )
+    }
+  }
+
+  <ErrorMessageExample />
+```
--- a/src/atoms/ErrorMessage/index.js
+++ b/src/atoms/ErrorMessage/index.js
@@ -1,0 +1,54 @@
+/* eslint-disable react/prop-types */
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import delay from 'lodash/delay';
+import TransitionGroup from 'react-transition-group/TransitionGroup';
+
+import Text from 'atoms/Text';
+import styles from './error_message.module.scss';
+
+const hocErrorMessage = BaseComponent => props =>
+  <TransitionGroup>
+    { props.condition && <BaseComponent {...props} /> }
+  </TransitionGroup>;
+
+
+class ErrorMessage extends Component {
+  constructor() {
+    super();
+    this.state = { visible: false };
+  }
+
+  componentDidEnter() {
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -91,8 +91,8 @@
     "react-dom": "^15.5.4",
     "react-hot-loader": "^1.3.1",
     "react-modal": "^1.7.7",
-    "react-stickynode": "^1.3.1",
```
drewdrewthis: You removed stickynode?

src/atoms/Transition/__tests__/transition.spec.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { expect } from 'chai';
+import { mount } from 'enzyme';
+
+import Transition from 'atoms/Transition';
```
drewdrewthis: If you think this needs more testing, please make a ticket in CH

src/molecules/formfields/DateField/index.js
```diff
@@ -41,9 +46,18 @@ function DateField(props) {
...
-        <div className={styles['error']}>{ meta.error }</div>
-      }
+
+      <Transition
+        condition={!!message}
+      >
+        <ErrorMessage>
+          <Text
+            type={7}
+          >
```
drewdrewthis: Why aren't you just passing in the message as a prop?

drewdrewthis: ^^ Same for other input fields

src/templates/Navigator/index.js
```diff
@@ -53,16 +52,9 @@ function Navigator(props) {
                   steps={stepProgressData}
                 />
 
-                <Sticky
```
drewdrewthis: Did you mean to remove the sticky? Or did you only take it out for the symlink and forget to put it back?

jmcolella: yes my bad

src/molecules/formfields/DateField/index.js
jmcolella: because that seems a bit over engineered. pretty easy to just wrap the children

package.json
jmcolella: it's coming back

src/atoms/Transition/__tests__/transition.spec.js
jmcolella: OK

jmcolella:
@drewdrewthis let me know your thoughts

src/atoms/ErrorMessage/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+            onClick={this.onClick}
+          >
+            Click Me to reveal error message
+          </Button>
+
+          <Transition
+            condition={this.state.visible}
+            component={ErrorMessage}
+          >
+            'Error'
```
drewdrewthis: Why do you need to pass in the children here? I looked at the component and I don't really get it. Are you passing in a component that then wraps itself around the children in the `Transition` component?

jmcolella:
@drewdrewthis take a look at [this](https://github.com/policygenius/athenaeum/pull/195/commits/b4eaaa640f6f199fe130e7ce0906dd63547d604c) commit for changes. I brought `TransitionGroup` into `ErrorMessage` and created an hoc type thing

src/atoms/ErrorMessage/index.js
```diff
@@ -0,0 +1,54 @@
...
+   * the message to be displayed for the error or warning
+   */
+  message: PropTypes.string,
+  /**
+   * the condition for which the message will be shown or hidden
+   */
+  condition: PropTypes.bool,
+};
+
+export default hoc(ErrorMessage);
```
drewdrewthis: Will this then be named "`hoc`"?

src/molecules/formfields/CreditCardField/index.js
```diff
@@ -48,13 +52,11 @@ function CreditCardField(props) {
...
       </div>
-      { meta && meta.touched && meta.error &&
-        <div className={styles['error']}>{ meta.error }</div>
-      }
 
-      {meta && meta.touched && meta.warning &&
-        <div className={styles.error}>{ meta.warning }</div>
-      }
+      <ErrorMessage
+        condition={!!message}
```
drewdrewthis: Seems like condition is unnecessary now, no?

src/atoms/ErrorMessage/index.js
jmcolella: not necessarily; when importing, you can name the module whatever you want, so I've used it as `ErrorMessage`

src/molecules/formfields/CreditCardField/index.js
jmcolella: the condition is necessary for `TransitionGroup` to fire its lifecycle methods. it's all dependent on a component being added/removed from the DOM, and the condition is what controls that

drewdrewthis: I meant that the existence of the error message is the conditional. Passing it twice seems redundant to me.

jmcolella: 👍

PR #197, opened 2017-05-26
HeaderAmount componenet added, Renamed PriceDiscountHeader, Add mobil…

patrickkim:
…eHeader spot


plop_templates/js_template.txt
```diff
@@ -9,15 +9,13 @@ function {{properCase name}}(props) {
     className,
   } = props;
 
+  const classes = [
+    styles['{{dashCase name}}'],
+    className,
+  ];
```
drewdrewthis: Nice. This is a good convention.

src/molecules/HeaderDiscount/header_discount.module.scss
```diff
@@ -12,18 +10,22 @@
...
-  vertical-align: super;
+  font-size: 60%;
+  line-height: inherit;
 }
 
-.sm {
-  @include typography-5(false);
+.curr {
+  position: relative;
+  vertical-align: 10%;
```
drewdrewthis: @patrickkim This seems like all repeated styles? Is there a way to give them a single source of truth?

patrickkim: true, not sure where they will live yet.

PR #200, opened 2017-05-31
Remove composes in our scss.

patrickkim:


src/atoms/Button/buttons.module.scss
```diff
@@ -90,7 +91,7 @@ $button-solid-background-tertiary-color:   #f08f5c;
 }
 
 .disabled {
-  composes: button;
+  @extend %button;
```
drewdrewthis: I know we discussed doing it this way, but do you think it might make sense to just actually give all the buttons the `.button` class, instead of extending this?

patrickkim: That's a great idea. Didn't think of that. standby.

PR #203, opened 2017-05-31
Add jest snapshots

drewdrewthis:


plop_templates/test_template.txt
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 //import sinon from 'sinon';
-import { expect } from 'chai';
 import { mount, shallow } from 'enzyme';
+import renderer from 'react-test-renderer';
```
jmcolella: why do we need this?

src/atoms/Layout/__tests__/layout.spec.js
```diff
@@ -1,22 +1,29 @@
...
-  it('renders', () => {
+  it('creates a Layout component', () => {
     const wrapper = mount(<Layout />);
 
-    expect(wrapper.type()).to.equal(Layout);
+    expect(wrapper.type()).toEqual(Layout);
+  });
+
+  it('renders correctly', () => {
+    const component = renderer.create(<Layout />).toJSON();
```
jmcolella: just curious why we need this `renderer` here. you don't seem to use it for the other snapshot checks

drewdrewthis: @jmcolella Not entirely sure the difference -- but I think it's similar to `mount` vs `shallow` in terms of what actually gets rendered into the snapshot

PR #206, opened 2017-05-31
Break out TableOfContent into seperate sectioned files

patrickkim:


patrickkim:
I think noobs will appreciate the longer name.

PR #209, opened 2017-06-01
Create TextArea form field component

trevornelson:


trevornelson:
@policygenius/mauve cc: @patrickkim @drewdrewthis 👀 

Text area component for use in life web ( https://github.com/policygenius/life-web/pull/67 )

src/molecules/formfields/TextArea/index.js
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * The props under the meta key are metadata about the state of this field that `redux-form` tracks.
+   */
+  meta: PropTypes.object,
+
+  /**
+   * Passing `noBaseStyle=true` will omit the base class style
+   */
+  noBaseStyle: PropTypes.bool,
```
drewdrewthis: hmm - this is interesting. @trevornelson Is this necessary?

trevornelson: @drewdrewthis Maybe not, I saw it in a few other components and thought it was a convention for the RCL. I can definitely remove, as I was really just thinking of this as boilerplate.

src/molecules/formfields/TextArea/text_area.module.scss
```diff
@@ -0,0 +1,25 @@
+@import '../shared/formfields.module.scss';
```
drewdrewthis: This file is bringing in a bunch of different classes, including the header classes. @patrickkim -- I thought we changed this? I think that this will create a new classname for each classname brought in from formfield.

src/molecules/formfields/TextArea/index.js
drewdrewthis: Ah, hm. @patrickkim @jmcolella -- is this a new convention?

jmcolella: @drewdrewthis @trevornelson we are using `noBaseStyle` for input fields that are being wrapped up as a collection. Ex: `DateField`. @cisacke introduced it as a way to git rid of some stuff like `border` and other styles so that the `DateField` wrapper would control the styling of those individual inputs as a group. 

Also see `CreditCardField`. I would suggest limiting the `noBaseStyle` prop to those types of situations

patrickkim: This conversation brings up a good point. We should provide better documentation as to what this prop is for and why. I noticed that in all our other input type fields we have this prop, so we should include it in along with this.

patrickkim: I'll add in the extra note on the `noBaseStyle` prop

src/molecules/formfields/TextArea/text_area.module.scss
```diff
@@ -0,0 +1,25 @@
...
+  padding-right: ru(.5);
+  border-bottom: $border;
+}
+
+textarea.textarea-field {
+  border-style: none;
+  border-color: Transparent;
+  overflow: auto;
+  outline: none;
+  padding: .63rem 1.125rem;
```
jmcolella: small comment, try to use `rem-calc` or `ru` if in increments of `.25`

src/molecules/formfields/TextArea/text_area.module.scss
patrickkim: I introduced this pattern with the FormField refactor. To move forward include this for now. 
I'll look into finding a permanent solution `@import formfield.module`

PR #212, opened 2017-06-01
Remove global stylesheet import

patrickkim:


patrickkim:
It easier to read the changes separately. Once you guys are a cool with this I'll squash it.

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -54,39 +47,7 @@ $focused-border-color: color('brand-3');
...
-.tooltip-wrapper {
-  position: absolute;
-  top: ru(.75);
-  right: ru(.75);
-}
-
-
-/* -- State:focus  -- */
+/* --  state:focus  -- */
 .focused {
```
patrickkim: I left .focused, .error classes in shared, due to the fact that they are used across all field components. It isn't a large ruleset so I think that we should be ok.

drewdrewthis: @patrickkim You could nest the classes in the placeholder, I believe -- it's not much of a difference, but it will scope it to whatever uses the placeholder only, enforcing our convention.

src/molecules/formfields/shared/formfields.module.scss
```diff
@@ -112,5 +72,5 @@ $focused-border-color: color('brand-3');
 }
 
 .hasError {
-  border: 1px solid $field-error-color;
+  border: 1px solid color('accent-2');
```
drewdrewthis: Since this is a shared color / standard for the component, I think it makes sense to enforce the consistency via a variable, no?

src/organisms/cards/RadioCard/index.js
```diff
@@ -41,16 +41,16 @@ function RadioCard(props) {
...
     <div
-      className={
-        classnames(
-          styles['radio-card'],
-          { [styles.checked]: input.value === radioValue },
-          { [styles.sectioned]: isValid(sections) },
-          className
-        )
-      }
+      className={classnames(...classes)}
```
drewdrewthis: Why did you decide not to include the `classnames` method as well into the `classes` constant? ie:

`const classes = classnames( styles['radio-card'], ...)`

patrickkim: Style preference really. No real upor downside to it other than I like using spread operator. Moving forward, we should use `const classes = classnames(...`

src/molecules/formfields/shared/formfields.module.scss
patrickkim: I was avoiding doing this in fear of breaking things on Renters but I went in and did another css audit. I think I have the shared fields slimmed down their barebone minimum. I tested things out and it all seems to work as expected.

PR #214, opened 2017-06-02
Make sidebar in Checkout sticky

cisacke:


cisacke:
@drewdrewthis @jmcolella @patrickkim should be a really quick 👀  if someone has a minute!

PR #222, opened 2017-06-06
Create latest directory

drewdrewthis:
Background: 
- The RCL simulates a routing system, so it needs a consistent "root" directory.
- It is convenient to be able to go to the latest version without having to know what the version number is.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -27,7 +27,7 @@ else
   VERSION="$BUILDKITE_BRANCH-$BUILDKITE_BUILD_ID"
 fi
 
-echo '+++ Deploying styleguide to S3'
+echo '+++ Deploying version-specific styleguide to S3'
 aws s3 sync \
   ./styleguide s3://rcl-pg/$VERSION/ \
   --acl=public-read \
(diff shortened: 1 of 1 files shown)


bin/deploy
```diff
@@ -36,5 +36,14 @@ aws s3 sync \
 
 echo "Deployed styleguide to http://rcl.policygenius.com/$VERSION/index.html"
 
+echo '+++ Deploying styleguide latest to S3'
+aws s3 sync \
+  ./styleguide s3://rcl-pg/latest/ \
```
drewdrewthis: @j-clark My only concern is that I might have to force the overwriting to this directory. I'm not sure ~

j-clark: problem 1: you're telling browsers to cache the result of this page for a very long time, which means they won't be getting the latest version latest
problem 2: this S3 bucket has a cloudfront distro in front of it and we'd need to manually invalidate their cache (which costs money and is slow) every time we deploy a new thingy

if you resolve those 2 problems, it'd probably work

drewdrewthis: @j-clark The cache thing is a one-liner in the deploy script, I believe. #2.. I'm not sure. But would like to discuss further with you offline

PR #223, opened 2017-06-06
Tweaks for playback card, checkout & general form

cisacke:
for AC on [this ticket](https://www.pivotaltracker.com/story/show/146473497)


cisacke:
@policygenius/front-end 👀  please!

src/organisms/cards/PlaybackCard/index.js
```diff
@@ -14,10 +15,11 @@ function PlaybackCard(props) {
     onClick,
     icon,
     editLinkText,
+    marginBottom,
   } = props;
 
   return (
-    <div className={styles['playback-card']}>
+    <div className={classnames(styles['playback-card'], marginBottom && styles['add-margin'])}>
```
drewdrewthis: @cisacke Components should not have their own margins -- they should be positioned by their parents. In the future, it will be hard / complicated to reposition this Playback component in a different context.

src/organisms/cards/PlaybackCard/playback_card.module.scss
```diff
@@ -77,14 +77,21 @@
...
   }
 
   .amount {
     text-align: center;
     margin-top: ru(.5);
   }
 }
+
+@media #{$medium-up} {
+  .add-margin {
```
drewdrewthis: Same comment as above.

src/organisms/forms/GeneralForm/index.js
```diff
@@ -8,19 +8,11 @@ function GeneralForm( props ) {
...
   return (
-    <div className={classnames( styles[''], className )}>
-      <form
-        onSubmit={onSubmit}
-        name={formName}
-      >
-        { children }
-      </form>
+    <div className={classnames( styles['general-form'], className )}>
+      { children }
```
drewdrewthis: Do we not want the `GeneralForm` to create a form? Why not just remove the component?

src/organisms/cards/PlaybackCard/index.js
cisacke: @drewdrewthis in this case I can't position from the parent because the design requires the margin only on medium-up screen sizes ([design](https://app.zeplin.io/project/58e6498f59f26ca94d348983/screen/590c9b1082a395bab0c9e57e)) - the other way I thought I might be able to solve this was to change the `Spacer` component to take a prop that would allow for the spacer to only show on certain screen sizes - what do you think?

src/organisms/forms/GeneralForm/index.js
drewdrewthis: @cisacke What benefit are we gaining with this form component? Should we just remove it?

src/organisms/cards/PlaybackCard/index.js
drewdrewthis: Hmm -- how about a component that handles hiding for us? Either a wrapper or a HOC that can take a `hideOn` string prop: `hideOn='small large'`

src/organisms/forms/GeneralForm/index.js
patrickkim: Alternatively, can't you just pass in the form props on `<GeneralForm />` and spread them on them on `<form />`  with `...rest`

cisacke: @patrickkim @drewdrewthis the only benefit is [the css](https://github.com/policygenius/athenaeum/pull/223/files#diff-85c27aa18b6dc3064078588be7297e3dR38), which is adding horizontal padding on small screen sizes in order to accommodate [this design](https://app.zeplin.io/project/58e6498f59f26ca94d348983/screen/590c9b1082a395bab0c9e57e), which requires the individual components to set their own padding instead of the checkout container. I can't inline this because it's a media query, and I don't really want to write a wrapper component because it seems to be more of a unique use case than, for example, the `Hide` component I just wrote. The only other thing I can think of is to introduce a stylesheet to renters -- what are your thoughts?

drewdrewthis: @cisacke Sounds like a variant or a new prop that we can add to `Layout`?

cisacke:
@drewdrewthis thanks for your input on all of this, it's been super useful -- I've implemented two of your suggestions - wrote a [`Hide` component](https://github.com/policygenius/athenaeum/pull/223/commits/a65bfd6c8cf6da50e4d9a636dcfa151f393c2f7e) and [added a variant to `Layout`](https://github.com/policygenius/athenaeum/pull/223/commits/b27d30b562ce96492c18af66ce84672a670c6c93) to handle the padding issue

src/wrappers/Hide/index.js
```diff
@@ -0,0 +1,45 @@
...
+Hide.propTypes = {
+  /**
+   * This prop will add a new className to any inherent classNames
+   * provided in the component's index.js file.
+   */
+  className: PropTypes.string,
+  /**
+   * The screen sizes on which to hide the children -- use screen sizes from `Layout` component
+   */
+  hideOn: PropTypes.string.isRequired,
```
patrickkim: It would be nicer and more consistent if this also supported boolean value for sizes in addition to className. It would help with the documentation of this component as well.

src/atoms/Layout/index.js
```diff
@@ -124,7 +124,7 @@ Layout.propTypes = {
    *
    * `'bordered-buckets'`: Creates borders around `Col` from medium viewports up.
    */
-  variant: PropTypes.oneOf([ 'bordered-buckets' ]),
+  variant: PropTypes.oneOf([ 'bordered-buckets', 'padding-small' ]),
```
patrickkim: This seems hyper specific but given the problem you need to solve at hand, this seems like the easiest way to solve it.

src/wrappers/Hide/index.js
cisacke: @patrickkim that's a good idea - I'm going to merge as is so that I can unblock these renters tickets but I'll go back and make that change later

src/atoms/Layout/index.js
cisacke: 👍

src/atoms/Layout/layout.module.scss
```diff
@@ -40,6 +40,16 @@ $border: 1px solid color('neutral-6');
   }
 }
 
+@media #{$small-only} {
+  .padding-small {
```
patrickkim: I think it would be helpful to add a little comment here & where it's used. Eventually I would like to see this deprecated and a more robust way to handle these types of use cases.

PR #226, opened 2017-06-08
Add tooltip to TextField label & CreditCardInput CCV

cisacke:
For AC on [this ticket](https://www.pivotaltracker.com/story/show/145950051) - basically just need to allow for a tooltip in both the `TextField` input & label that takes onClick handler. I also added the ability to make it a regular tooltip with text, since that seems a likely use case in the future.

### DIFF
--- a/src/atoms/Tooltip/index.js
+++ b/src/atoms/Tooltip/index.js
@@ -35,11 +35,12 @@ class Tooltip extends React.Component {
       className,
       left,
       right,
+      onClick,
     } = this.props;
 
     return (
       <span
-        onClick={this.openModal}
+        onClick={onClick || this.openModal}
         className={classnames(
           styles['tooltip-wrapper'],
           className
@@ -48,7 +49,7 @@ class Tooltip extends React.Component {
         <Icon icon='questionMark' className={styles['tooltip']} />
         <span
           className={classnames(
-            styles['hover-message'],
+            children && styles['hover-message'],
             left && styles['left'],
             right && styles['right'],
           )}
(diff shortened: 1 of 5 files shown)


cisacke:
@drewdrewthis @patrickkim @jmcolella 👓  please -- also some other PRs [here](https://github.com/policygenius/athenaeum/pull/224) and [here](https://github.com/policygenius/athenaeum/pull/223) that are blocking renters, if anyone has a minute - thanks!

src/atoms/Tooltip/index.js
```diff
@@ -35,11 +35,12 @@ class Tooltip extends React.Component {
       className,
       left,
       right,
+      onClick,
     } = this.props;
 
     return (
       <span
-        onClick={this.openModal}
+        onClick={onClick || this.openModal}
```
drewdrewthis: Smart.

src/molecules/formfields/TextField/index.js
```diff
@@ -125,6 +140,20 @@ TextField.propTypes = {
    * id added to the `input` node
    */
   id: PropTypes.string,
+  /**
+   * either a handler for clicking the tooltip, or text to go in the tooltip for the label
+   */
+  tooltip: PropTypes.oneOfType([
```
drewdrewthis: Cool

src/molecules/formfields/TextField/index.js
```diff
@@ -2,12 +2,22 @@ import React from 'react';
...
+import isString from 'lodash/isString';
 
 import Icon from 'atoms/Icon';
+import Tooltip from 'atoms/Tooltip';
 import ErrorMessage from 'atoms/ErrorMessage';
 
 import styles from './text_field.module.scss';
 
+function renderTooltip(tooltip, className) {
+  if (isString(tooltip)) {
```
drewdrewthis: Do we need to lean on the library for this instead of just writing on the js?

cisacke: @drewdrewthis yeah good point I was being lazy - will write out the js

PR #227, opened 2017-06-08
Fix navigator layout for mobile

jmcolella:


jmcolella:
@cisacke merging based on our debugging

PR #228, opened 2017-06-12
Add Error messages to Toggle Fields

jmcolella:
@cisacke Adds error message to Toggle Fields for knockout questions


cisacke:
@jmcolella just double check there aren't any linting errors - I got a few locally

PR #229, opened 2017-06-14
Mobile calculators are sticky

cisacke:
this should be everything needed to unblock [this ticket](https://www.pivotaltracker.com/story/show/145686023) but I couldn't check locally because of the sticky issue, so might need some tweaks


cisacke:
@jmcolella quick 👀  please, when you have a minute

cisacke:
@jmcolella think I made a [small improvement](https://github.com/policygenius/athenaeum/pull/229/commits/d4abe6a5a09fad41e35f0e1c2586045cb90a5798) with the icon sizing - it's not perfect & on super small screens it's still tiny but I didn't want to spend too much time on it

PR #232, opened 2017-06-14
Add smart media query specific linebreak element

patrickkim:


patrickkim:
@cisacke @jmcolella @drewdrewthis CR please.

PR #235, opened 2017-06-15
fix Sticky userAlert, change headerAmount

cisacke:
- fixes padding on checkout mobile
- fixes user alert on checkout to work with existing Sticky component
- tweaks the checkout header to allow for more text in the `unit` prop
- fixes the sidebar Sticky `top` prop by accounting for `userAlert.condition`

### DIFF
--- a/src/templates/CheckOut/index.js
+++ b/src/templates/CheckOut/index.js
@@ -8,6 +8,33 @@ import UserAlert from 'atoms/UserAlert';
 import Text from 'atoms/Text';
 import styles from './checkout.scss';
 
+const CheckoutUserAlert = ({ userAlert, enabled, className }) =>
+  <UserAlert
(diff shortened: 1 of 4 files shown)


cisacke:
@jmcolella 👀  please

src/templates/CheckOut/index.js
```diff
@@ -76,40 +101,34 @@ function CheckOut(props) {
                   <Col className={styles['cost-price']}>
                     { curr && <sup>{curr}</sup> }
                     { value }
-                    { unit && <small>/{unit}</small> }
+                    { unit && <small>{unit}</small> }
                   </Col>
                 </Layout>
               </Col>
+              { userAlert.condition &&
```
jmcolella: smart, nice 👍

PR #236, opened 2017-06-15
Fix some css issues with mobile modal

cisacke:


cisacke:
@jmcolella was going to wait on some icons from natalie but this can be reviewed as well -- thanks!

src/molecules/Modal/modal.module.scss
```diff
@@ -24,6 +24,14 @@ $border: 1px solid color('neutral-5');
       padding: 0;
       border-bottom: none;
     }
+
+    .section:not( :first-child ) {
```
jmcolella: I know we're trying to avoid using `not`, but I don't think there's a better way to do this, unless we use `> *` which we are also trying to avoid

drewdrewthis: @jmcolella Unless I'm missing something, you could have just updated the `.section` props, and then have `:first-child` override?

PR #239, opened 2017-06-16
apiKey should be key

cisacke:
static maps is expecting the apiKey to be `key` - also happens to be a reserved prop name which is why I didn't change the prop to `key`


cisacke:
@jmcolella 👀  please

PR #240, opened 2017-06-16
Add icons to radio card

jmcolella:
@cisacke CR please

### DIFF
--- a/src/organisms/cards/RadioCard/index.js
+++ b/src/organisms/cards/RadioCard/index.js
@@ -3,8 +3,10 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { camelCase, omit } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


src/organisms/cards/RadioCard/index.js
```diff
@@ -123,6 +147,12 @@ RadioCard.propTypes = {
   image: PropTypes.string,
 
   /**
+   * optional icon displayed below description; `icon` overrides `image` prop;
+   * This is the icon name from the [Icon component](/#icon).
+   */
+  icon: PropTypes.string,
```
cisacke: we might also want to allow the user to pass in the height and width props for the icon component?

jmcolella: @cisacke ok, I added `iconProps` object that spreads over `Icon`

PR #243, opened 2017-06-19
Add markers and subheader to StaticMap

cisacke:


cisacke:
@jmcolella 👀  please - need this for the zip code renters ticket. not sure where the snapshot for `Br` is coming from but looks like it never got committed 🤷‍♀️

src/organisms/cards/StaticMapCard/Readme.md
```diff
@@ -4,6 +4,7 @@ StaticMapCard Example:
     <StaticMapCard
       label='A map has no label'
       center='10010'
+      subheader='New York, NY 10011'
```
jmcolella: just curious how this works in implementation? are we going to ping google for this?

cisacke: @jmcolella yes in renters we are gonna hit their geocode api to reverse geocode the zip, and then pass that value to `subheader`

PR #244, opened 2017-06-20
change unit font size and remove onClick from select field input

cisacke:
@jmcolella 👀  please -- I noticed a bug in renters where you can't change the ppc select card field because it opens the modal when you click on the select field. I think this is because `onClick` is getting passed to both `SelectCard` and `SelectField` through `input`, so the whole field is clickable. Couldn't get symlink to work because of the react-modal issue, but I think this will fix it. Also changed the font size of something on the checkout as part of AC on [this ticket](https://www.pivotaltracker.com/story/show/145686023)

### DIFF
--- a/src/molecules/formfields/SelectField/index.js
+++ b/src/molecules/formfields/SelectField/index.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import omit from 'lodash/omit';
 
 import Tooltip from 'atoms/Tooltip';
 import ErrorMessage from 'atoms/ErrorMessage';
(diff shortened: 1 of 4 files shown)


src/templates/CheckOut/checkout.scss
```diff
@@ -72,6 +72,10 @@ $mobile-layout-padding: ru(1);
 
   &-price {
     display: none;
+
+    small {
```
jmcolella: should we just use `Text` component for this rather than the `small` html node? just seems like an anti-pattern to use an arbitrary percentage for font size rather than our brand sizes

cisacke: @jmcolella good call, changed

PR #246, opened 2017-06-21
fix navigator height issue

cisacke:


cisacke:
@jmcolella 👀  please - should fix the height issue by keeping the container `display: flex` on medium-up, and `display: initial` on small screens to prevent that other bug we were running into on mobile

jmcolella:
@cisacke LGTM to me

PR #262, opened 2017-07-13
Update Modal and add rcl- prefix to classnames

drewdrewthis:


src/molecules/Modal/index.js
```diff
@@ -120,7 +128,8 @@ Modal.propTypes = {
 };
 
 Modal.defaultProps = {
-  variant: 'simple'
+  variant: 'simple',
+  hideIcon: false
```
jmcolella: is this supposed to be `hideX`?

drewdrewthis: Yup!

PR #266, opened 2017-07-18
Fix safari mobile issues

jmcolella:


jmcolella:
@cisacke CR please. 

@drewdrewthis would also like your eyes on this

src/templates/CheckOut/checkout.scss
```diff
@@ -5,7 +5,7 @@ $mobile-layout-padding: ru(1);
   width: 100%;
   min-height: 100vh;
   background-color: color('neutral-7');
-  display: flex;
+  // display: flex;
```
cisacke: need this?

src/templates/CheckOut/checkout.scss
```diff
@@ -182,7 +187,7 @@ $mobile-layout-padding: ru(1);
   .sticky {
     @include sticky-disable;
 
-    height: 100% !important;
+    height: 77px !important;
```
cisacke: maybe extract this into a constant so there's some context around the number?

PR #275, opened 2017-07-24
CSS padding tweaks

cisacke:
@jmcolella 👀 please


src/molecules/Modal/modal.module.scss
```diff
@@ -87,8 +87,14 @@ $border: 1px solid color('neutral-5');
 }
 
 @media #{$small-only} {
-  .modal {
-    padding: 1rem;
+  .modal, .section, .header {
+    padding: ru(1);
```
jmcolella: is this value following the Zeplins? PT says padding should be smaller on mobile, but this increases it

cisacke: @jmcolella yeah from the zeplins - it is decreasing on mobile but the relevant classes are the `.section` and `.header`

PR #280, opened 2017-07-26
Update font color for fields

jmcolella:


jmcolella:
@cisacke similar PR as before but i refactored and updated color for `invalid` select pseudo selector

PR #285, opened 2017-08-09
change onBlur handler for cc field

cisacke:
@jmcolella @bradwheel this is my proposed solution for [this ticket](https://app.clubhouse.io/policygenius/story/4941/user-shouldn-t-see-invalid-state-on-cc-payment-info-until-they-ve-clicked-out-of-last-form-field).

Basically this just makes sure that the `onBlur` event only gets fired when the target is outside of the input in order to mimic how the event should get fired, and allows us to use the `meta.active` property correctly to display or hide validations

### DIFF
--- a/src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
+++ b/src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
@@ -7,7 +7,7 @@ import CreditCardField from 'molecules/formfields/CreditCardField';
 describe('<CreditCardField />', () => {
 
   it('renders', () => {
-    const wrapper = mount(<CreditCardField input={{}} />);
+    const wrapper = mount(<CreditCardField input={{}} meta={{}} />);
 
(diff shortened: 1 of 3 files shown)


src/molecules/formfields/CreditCardField/__tests__/credit_card_field.spec.js
```diff
@@ -7,7 +7,7 @@ import CreditCardField from 'molecules/formfields/CreditCardField';
 describe('<CreditCardField />', () => {
 
   it('renders', () => {
-    const wrapper = mount(<CreditCardField input={{}} />);
+    const wrapper = mount(<CreditCardField input={{}} meta={{}} />);
```
jmcolella: I think you also need to add `meta` filed in the `Readme.md`. I got an error when I spun it up locally

cisacke: @jmcolella yup good catch, added

PR #294, opened 2017-09-01
Fix search icon for swiftype mobile

jmcolella:


jmcolella:
@patrickkim merging this fix

PR #298, opened 2017-09-05
Turn mangle false

jmcolella:


jmcolella:
@patrickkim 

- Removes all animations due to issue with CSS Modules processing

Left a chore in the Main Nav backlog to fix at some point

PR #300, opened 2017-09-07
Add policy card

cisacke:
https://app.clubhouse.io/policygenius/story/6481/pg-engineers-should-have-a-new-policy-card-component


src/organisms/cards/PolicyCard/PolicyActions.js
```diff
@@ -0,0 +1,85 @@
...
+import accounting from 'accounting';
+import classnames from 'classnames';
+import Text from 'atoms/Text';
+import Layout from 'atoms/Layout';
+import Button from 'atoms/Button';
+import Hide from 'wrappers/Hide';
+import Spacer from 'atoms/Spacer';
+import styles from './policy_card.module.scss';
+
+const ResponsiveText = ({ text, offset = 0 }) =>
```
cisacke: this is super janky but basically bc of the layout, the buttons get super small when changing the width of the browser and this was the only way I could get the font size to respond to the width of the buttons

src/organisms/cards/PolicyCard/propTypes.js
```diff
@@ -0,0 +1,61 @@
...
+   */
+  monthlyPremium: PropTypes.number.isRequired,
+  /**
+   * Optional discount percentage
+   */
+  discount: PropTypes.node,
+  /**
+   * Optional array of nodes that will be spaced appropriately in the footer
+   */
+  footer: PropTypes.array,
```
cisacke: my thinking here (and for `policyTooltip`) is that the icon will have to be styled and aligned during the implementation. This is because all icons are slightly different, and there's not one consistent way to change the color of a svg or set the vertical alignment

jmcolella: @cisacke sounds good to me

src/molecules/formfields/CheckBoxField/Readme.md
```diff
@@ -1,7 +1,7 @@
 Checkbox Example:
 
 ```example
-  <CheckboxField
+  <CheckBoxField
```
jmcolella: @cisacke was this intentional?

src/organisms/cards/PolicyCard/CarrierLogo.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Image } from 'cloudinary-react';
+import styles from './policy_card.module.scss';
+
+export const CarrierLogo = ({ carrierLogo }) =>
+  <div className={styles['carrier-logo']}>
+    <Image
```
jmcolella: @cisacke are we going to be using cloudinary to host these images? seems like they are being served up by rails, at least that's how it is in development

src/organisms/cards/PolicyCard/Compare.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Text from 'atoms/Text';
+import Hide from 'wrappers/Hide';
+import Spacer from 'atoms/Spacer';
+import CheckBoxField from 'molecules/formfields/CheckBoxField';
+import styles from './policy_card.module.scss';
+
+export const Compare = ({ compareSelected, onCompare }) =>
```
jmcolella: do we need this `export`?

src/molecules/formfields/CheckBoxField/Readme.md
cisacke: @jmcolella ahh yeah I just added a comment in the pr description about this - there's a discrepancy on github with the filename which was causing buildkite to fail

src/organisms/cards/PolicyCard/Compare.js
```diff
@@ -0,0 +1,33 @@
...
+  <Hide hideOn='small' className={styles['compare']}>
+    <Text
+      color='neutral-3'
+      light
```
jmcolella: @cisacke will this always stay `light`, even when selected?

src/organisms/cards/PolicyCard/PolicyActions.js
```diff
@@ -0,0 +1,85 @@
...
+        </Button>
+        <Button onClick={onDetails}>
+          <Hide hideOn='small'>
+            <ResponsiveText text='Details' offset={-10} />
+          </Hide>
+          <Hide hideOn='medium large xLarge'>
+            Details
+          </Hide>
+        </Button>
+        <Hide hideOn='medium large xLarge'>
```
jmcolella: @cisacke you need to add `xxLarge` here

src/organisms/cards/PolicyCard/PolicyActions.js
jmcolella: @cisacke I think we should push back on this design for screens between 769-1024. It's not even a matter of these texts, the buttons becoming that small is pretty ridiculous

src/organisms/cards/PolicyCard/PolicyInformation.js
```diff
@@ -0,0 +1,44 @@
...
+      <div className={styles['label']}>Customer Service</div>
+      <Text color='neutral-2' semibold>{customerService}</Text>
+    </Text>
+    <Text
+      type={7}
+      color='neutral-3'
+      semibold
+    >
+      <div className={styles['label']}>Total Customers</div>
+      <Text color='neutral-2' semibold>{accounting.formatNumber(totalCustomers)}</Text>
```
jmcolella: @cisacke maybe better to offload this formatting for implementation? what do you think?

src/organisms/cards/PolicyCard/CarrierLogo.js
cisacke: @jmcolella unless there's a reason not to, I think this should be ok? I implemented it this way because it's how we do it in life-web, and I didn't want to use `Icon` and have to add all the carriers as svgs

src/organisms/cards/PolicyCard/policy_card.module.scss
```diff
@@ -0,0 +1,127 @@
...
+
+  @if $centered {
+    align-items: center;
+  }
+}
+
+.policy-card {
+  border: 1px solid $border;
+  border-radius: ru(.25);
+  max-width: rem-calc(1060px);
```
jmcolella: @cisacke obviously this was intentional, I just don't see anything about `max-width` for the card in the zeplin, wanted to double check

src/organisms/cards/PolicyCard/Compare.js
cisacke: @jmcolella good q, not sure - I'll ask Jim

src/organisms/cards/PolicyCard/policy_card.module.scss
```diff
@@ -0,0 +1,127 @@
...
+    max-width: 140px;
+    margin: auto;
+    margin-bottom: ru(.5);
+    line-height: ru(1);
+  }
+}
+
+@media #{$medium-up} {
+  .compare {
+    @include section(2, true);
```
jmcolella: @cisacke you can probably simplify these `flex-grow` values.

it seems like `compare` and `carrier-logo` are same width while `type`, `info` and `actions` are the same. You could probably just have `flex-grow: 0` for the first two and then `flex-grow:1` for the last 3. this still ensures that the leftover space is distributed proportionally between these two sets of items as well as evenly within each set.

src/organisms/cards/PolicyCard/PolicyActions.js
patrickkim: Cool hack. Did you consider making this available as an actual atom.

patrickkim:
👍

src/organisms/cards/PolicyCard/PolicyInformation.js
cisacke: yeah good point - I think if there was uncertainty around the content of the information panel we'd want to keep the props as agnostic as possible but since the labels are hardcoded, if product decides to change anything we'd have to go back into the component and update. I'm gonna keep it as is for that reason but possible in the future we could condense the three props to an array of objects

src/organisms/cards/PolicyCard/policy_card.module.scss
cisacke: 👍 confirmed with Jim, we're going to have max-width of 1200px

src/organisms/cards/PolicyCard/policy_card.module.scss
cisacke: 👍 good call, I made that change

src/organisms/cards/PolicyCard/PolicyInformation.js
jmcolella: sounds good

src/organisms/cards/PolicyCard/CarrierLogo.js
cisacke: -- take this back, it does make sense to use the current implementation, I've made a PR for this [here](https://github.com/policygenius/athenaeum/pull/302)

PR #320, opened 2017-09-18
Add new button variant, add help cursor to PolicyType

cisacke:
@jmcolella 👀  please - what do you think about allowing multiple variants for Button? Maybe better to add a prop, `slim` which is boolean? Also just added some tweaks for [this ticket](https://app.clubhouse.io/policygenius/story/7380/shopper-should-be-able-to-click-details-button-on-comparison-page)

### DIFF
--- a/src/atoms/Button/buttons.module.scss
+++ b/src/atoms/Button/buttons.module.scss
@@ -107,6 +107,12 @@ $button-solid-background-tertiary-color:   #f08f5c;
   color: $button-info-color;
 }
 
+.slim {
+  padding-top: ru(.5);
+  padding-bottom: ru(.5);
+  line-height: 1rem;
+}
+
 .toggle {
   @include typography-7(false);
   @include typography-regular;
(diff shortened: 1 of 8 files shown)


src/atoms/Button/buttons.module.scss
```diff
@@ -107,6 +107,12 @@ $button-solid-background-tertiary-color:   #f08f5c;
   color: $button-info-color;
 }
 
+.slim {
+  padding-top: ru(.5);
+  padding-bottom: ru(.5);
+  line-height: 1em;
```
jmcolella: Is there a reason this isn't rem?

jmcolella:
@cisacke actually sorry, changed my mind. I think the Boolean is a better idea for slim since it's not really a variant, it's just an extra condition for a variant

src/atoms/Button/buttons.module.scss
cisacke: not sure, I just copied it from one of the other variants - I'll see if I can convert them both to rem

PR #322, opened 2017-09-19
The nav works as intended

patrickkim:


src/static/MainNav/index.js
```diff
@@ -63,7 +63,15 @@ class MainNav extends Component {
   }
 
   toggleMobileMenu = (toggle) => {
-    this.setState({ showMobileMenu: toggle });
+    setTimeout(() => {
```
patrickkim: Sorry everyone. :(

PR #328, opened 2017-09-21
Fix PolicyCard IE11

jmcolella:
@cisacke CR please

[CH](https://app.clubhouse.io/policygenius/story/7381/shoppers-should-see-correct-policycards-in-ie11-and-firefox)

- Fixes `PolicyCard` layout for IE 11 (and Firefox/Safari)
- Gives `max-width` to all `Layout` children

@drewdrewthis please weigh in on the second bullet point. IE has a bug where flex children do not respect `box-sizing` if there isn't a `max-width` set, essentially making all children of `Layout` broken (particularly those children that don't take up 12 columns). To rectify this, I added `max-width` to all children, equal to its `width` and `flex-basis`. Since we are setting `flex-grow: 0`, this change didn't seem to change much, but wanted to get your opinion. You can read more about the bug [here](https://github.com/philipwalton/flexbugs/issues/3)

### DIFF
--- a/src/atoms/Layout/layout.module.scss
+++ b/src/atoms/Layout/layout.module.scss
@@ -7,6 +7,7 @@ $spacer-base: .25;
 
   width: $width;
   flex-basis: $width;
+  max-width: $width;
   flex-grow: 0;
   flex-shrink: 1;
 }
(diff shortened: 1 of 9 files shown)


cisacke:
@jmcolella LGTM - agree that @drewdrewthis should weigh in on the layout thing. does it work on Firefox, too?

jmcolella:
@cisacke I checked all browsers, should be good to go

PR #333, opened 2017-09-22
WIP - Fix PolicyCard checkbox bug

sunnymis:
Added an onClick prop to the checkbox

### DIFF
--- a/src/molecules/formfields/CheckBoxField/index.js
+++ b/src/molecules/formfields/CheckBoxField/index.js
@@ -9,6 +9,7 @@ function CheckBoxField( props ) {
     label,
     input,
     fontStyle,
+    onClick
   } = props;
 
   return (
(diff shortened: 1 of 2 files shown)


jmcolella:
@sunnymis are you sure this is OK? I remember I removed the `onClick` from the checkbox itself because I delegated that functionality to the parent wrapper. I think having the `onClick` on the checkbox and the wrapper causes the function to fire twice and therefore not add the policy for comparison. Can you confirm that this isn't the case?

sunnymis:
@jmcolella Currently I am observing `onClick` firing twice in some instances. When a user clicks the checkbox it selects and deselects right away. Removing the `onClick` from the parent wrapper and adding it to the checkbox has solved this. But unfortunatley this has reduced the click coverage area, so it is still a WIP

PR #336, opened 2017-09-25
Fix Djanky Dropdowns in FF

danielnovograd:
- Fixes Select Cards & Filter Modules.


src/molecules/Filter/filter.module.scss
```diff
@@ -29,6 +29,19 @@
...
 
+@-moz-document url-prefix() {
+  .select-wrapper {
+    select {
+      -moz-appearance: none;
+    }
+    
+    select:-moz-focusring {
+      color: transparent;
+      text-shadow: 0 0 0 #4D4D4D;
```
cisacke: can we use `color('neutral-2')` here?

src/molecules/formfields/SelectField/select_field.module.scss
```diff
@@ -34,6 +34,20 @@
...
+@-moz-document url-prefix() {
+  .select-wrapper {
+    select {
+      -moz-appearance: none;
+      color: rgba(0,0,0,0);
+    }
+    
+    select:-moz-focusring {
+      color: transparent;
+      text-shadow: 0 0 0 #00a4a5;
```
cisacke: same here, I think this is brand-2

PR #355, opened 2017-10-06
Fix selects in firefox

PepperTeasdale:
We were passing a new random string to the key prop every time a select
option was rendered. For some reason Chrome knew how to deal with that,
but Firefox couldn't. This caused inputs to lose their selected value
when re-rendered.


src/utils/fieldUtils/renderOption.js
```diff
@@ -16,7 +14,7 @@ export default function renderOption(opt, idx = 0) {
 
   return (
     <option
-      key={reactKey ? `${reactKey}-${idx}` : `opt-${randomStr()}`}
+      key={reactKey ? `${reactKey}-${idx}` : idx.toString()}
```
danielnovograd: Nice!

PR #358, opened 2017-10-09
Update fonts

sunnymis:
Updated fonts to match the hosted ones. No longer using separate names, but differentiating based on weight


sunnymis:
@jmcolella 👁 👁

PR #360, opened 2017-10-12
Add styled components

drewdrewthis:


src/atoms/StyledWrapper/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'styled-components';
+import { breakpoints } from './breakpoints';
+
+function StyledWrapper({ children, styles, tag }) {
+  return React.createElement(
+    styles ? styled(tag)(css([ styles({ breakpoints }) ])) : tag,
+    {},
```
sunnymis: Would it make sense to pass props into here instead of empty object? This way we can conditionally apply styles based on a prop.

drewdrewthis: Yeah... but there's no good use case I could think of off the top of my head, since the consumer is writing out the styles string themselves. Figured when there was a specific need, we could update it. Let me know if you can think of one

PR #363, opened 2017-10-13
Main nav rebrand

jmcolella:


jmcolella:
@sunnymis @drewdrewthis CR please

- Adds a ton of icons
- Update `List` to opt out of padding and allow for hoverable list item links

PR #381, opened 2017-10-19
Add 13px and 16px spacing

sunnymis:
The new designs for lists have two main bottom spacings, 13px and 16px. Doesn't conform to the ru's but this has to be here. 

@drewdrewthis @jmcolella 👀

### DIFF
--- a/src/atoms/List/list.module.scss
+++ b/src/atoms/List/list.module.scss
@@ -1,6 +1,8 @@
  $size-map: (
   6: ru(.25),
   12: ru(.5),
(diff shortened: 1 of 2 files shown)


src/atoms/List/list.module.scss
```diff
@@ -1,6 +1,8 @@
  $size-map: (
   6: ru(.25),
   12: ru(.5),
+  13: 13px,
```
jmcolella: use `rem-calc` here so everything is in `rem`

PR #382, opened 2017-10-19
Typography line-height changes and bold fix

sunnymis:
- Matching line-heights exactly now. 
- Type B font bolding works correctly

@drewdrewthis @jmcolella 👀


jmcolella:
@sunnymis LGTM merging 4 u

PR #388, opened 2017-10-23
Add rebrand Add Card

jmcolella:


jmcolella:
@sunnymis @drewdrewthis CR please

PR #394, opened 2017-10-23
Add FeatuerdPolicyCard

jmcolella:


jmcolella:
@cisacke @danielnovograd @trevornelson CR please

[CH](https://app.clubhouse.io/policygenius/story/9185/pg-engineers-should-have-a-new-feature-card-component)

- Adds `FeaturedPolicyCard` for comparison page

src/organisms/cards/FeaturedPolicyCard/index.js
```diff
@@ -0,0 +1,219 @@
...
+          Continue
+        </Button>
+
+        <Spacer spacer={2} />
+
+        <LinkWrapper
+          onClick={onDetails}
+          className={styles['details-link']}
+        >
+          {'Deatils & Review'}
```
trevornelson: `Deatils` 😃

jmcolella:
@trevornelson yes we can certainly make the API a bit more flexible and not use such hardcoded values as `policyType` and `financialStrength`. This does open up the same question for the `PolicyCard`, which is just as hardcoded for `life-3` as this component.

I'm hesitant to rip this out of athenaeum and put it in life-3 as it kind of goes against working towards our UI being co-located. I understand the frustration with bumping small things in athenaeum, but maybe that's when we should be asking design/product for good reasons as to why some color should change, etc.

It's easy to make the API as flexible as possible and what we sacrifice in small UI bumps (like color), we gain in the organization/reusability of the components themselves.

jmcolella:
@trevornelson Please see my refactor [here](https://github.com/policygenius/athenaeum/pull/394/commits/3c9ae02a08a01574592088e95367d706f620fba2) based on our conversation

PR #405, opened 2017-10-27
Add mobile feature card

jmcolella:
@trevornelson @cisacke @danielnovograd CR please

[CH](https://app.clubhouse.io/policygenius/story/9306/shoppers-should-see-updated-mobile-comparison-page)

- Makes updates to `FeaturedPolicyCard` on mobile
- Changes mobile version of `PolicyCard` to be that of `FeaturedPolicyCard` and refactors props

### DIFF
--- a/src/organisms/cards/FeaturedPolicyCard/__tests__/__snapshots__/featured_policy_card.spec.js.snap
+++ b/src/organisms/cards/FeaturedPolicyCard/__tests__/__snapshots__/featured_policy_card.spec.js.snap
@@ -18,11 +18,11 @@ exports[`<FeaturedPolicyCard /> renders correctly 1`] = `
       >
         $19.80
          
-        <span
-          className="text typography-7 regular neutral-4 undefined span undefined"
+        <strong
+          className="text typography-7 semibold neutral-4 undefined strong undefined"
         >
           /MO
-        </span>
+        </strong>
       </strong>
       <div
         className="top-1"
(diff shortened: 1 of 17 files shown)


src/organisms/cards/FeaturedPolicyCard/__tests__/__snapshots__/featured_policy_card.spec.js.snap
```diff
@@ -59,7 +59,7 @@ exports[`<FeaturedPolicyCard /> renders correctly 1`] = `
               style={undefined}
             >
               <p
-                className="text typography-7 regular undefined undefined p undefined"
+                className="text typography-7 regular undefined undefined p undefined policy-info-label"
```
danielnovograd: Are these `undefined` classes supposed to be here?

drewdrewthis: They aren't. This happens when there isn't a guard clause preventing the class names from being added to the class names array.

danielnovograd: Gotcha. Thanks for the pro tip! 

Just an FYI @jmcolella, there are a whole lot more in this file than just this  one line.

jmcolella: @danielnovograd yeah I see. This looks like a bunch of components would have to be refactored to prevent this. I'll make a note of it for rebrand and see if they can pick it up

PR #413, opened 2017-10-31
Feature card tweaks

jmcolella:


jmcolella:
@cisacke this makes small tweaks to the `FeaturedPolicyCard` based on design comments

PR #427, opened 2017-11-07
Display active step only in progress bar

trevornelson:
👀  plz @cisacke @danielnovograd @jmcolella removes step names except for the active step from the progress bar to fix alignment issues. [ch9144](https://app.clubhouse.io/policygenius/story/9144/user-should-see-a-fixed-progress-bar)

### DIFF
--- a/src/molecules/StepIndicator/ProgressBarStep.js
+++ b/src/molecules/StepIndicator/ProgressBarStep.js
@@ -71,14 +71,12 @@ export default class ProgressBarStep extends Component {
   }
 
   render() {
-    const {
-      step
-    } = this.props;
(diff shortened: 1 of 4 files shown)


jmcolella:
@trevornelson I may be missing something, but the SASS here doesn't seem to include all of the stuff we worked on together to get the bar to work on mobile. Was this done in another PR that I don't see, or was removing all the other names accept for the active step another way to solve?

trevornelson:
@jmcolella there ended up still being some issues that I found in some other parts of the flow. I spoke to Jim and he said to just have the active step centered above the breadcrumbs, which ended up fixing things more comprehensively.

PR #451, opened 2017-11-16
Fix main nav

jmcolella:


jmcolella:
@drewdrewthis @sunnymis CR please

- Makes adjustments to MainNav for speed
- Consolidates static webpack configs

PR #468, opened 2017-11-21
 Move keyframe animations from global file to component

howers:


webpack.config.base.js
```diff
@@ -47,9 +47,10 @@ const baseRules = modulesName => [
       fallback: 'style-loader',
       use: [
         {
-          loader: 'css-loader?importLoaders=3',
+          loader: 'css-loader',
           options: {
             modules: true,
+            importLoaders: 3,
```
drewdrewthis: 👍

PR #474, opened 2017-11-22
Update step indicator for rebrand

trevornelson:
@cisacke @danielnovograd @jmcolella 👀 Fixes `StepIndicator` for [CH11041](https://app.clubhouse.io/policygenius/story/11041/rebrand-the-navigator-application). Also fixes the `AddCard` component to not force typography when passing in a component instead of text.


trevornelson:
FYI @drewdrewthis @patrickkim

PR #475, opened 2017-11-24
Make buttons inline block when its a tag

sunnymis:


src/atoms/Button/index.js
```diff
@@ -29,12 +30,13 @@ function Button( props ) {
     unflex && styles['unflex'],
   ];
 
-  if (href) {
+  if (href || linkAttrs) {
```
drewdrewthis: @sunnymis Why don't you just expose and spread `attr` in regardless? We only want the `<a>` if it's a link, right?

sunnymis: @drewdrewthis The idea was linkAttrs would be an object containing
```
{
  href: '',
  rel: '',
  target: ''
}
```
because href is an attribute as well. We wouldn't have to pass both href and linkAttr but just one object.

PR #496, opened 2017-11-29
Update build

drewdrewthis:


.buildkite/athenaeum-main-nav.yml
```diff
@@ -11,55 +11,48 @@ steps:
 
   - wait
 
+  - block: 'Deploy MainNav AND Footer to Staging?'
+    prompt: "Please review and approve your changes at: http://rcl.policygenius.com/staging/main_nav/index.html"
```
jmcolella: I think this prompt is unnecessary

drewdrewthis: It doesn't actually fire anyway

PR #501, opened 2017-11-30
Add modal

drewdrewthis:
This adds the modal to the Footer used for export. Had to upgrade react-modal and make some fixes.


src/organisms/cards/FeaturedPolicyCard/__tests__/__snapshots__/featured_policy_card.spec.js.snap
```diff
@@ -78,7 +78,6 @@ exports[`<FeaturedPolicyCard /> renders correctly 1`] = `
               >
                 <div />
               </span>
-              <noscript />
```
drewdrewthis: Not sure why this is happening.

PR #534, opened 2017-12-08
Allow contact fields to be optional on navigator and footer

samanthavholmes:


samanthavholmes:
This allows navigator/footer to be used without contact fields (like phone number)

src/templates/Navigator/index.js
```diff
@@ -123,9 +123,12 @@ function Navigator(props) {
                 </Sticky>
               </div>
             </Col>
-            <div className={styles['contact-card']}>
-              <ContactCard {...contactProps} />
-            </div>
+            {
+              contactProps &&
```
PepperTeasdale: Is there ever a chance that this will not be a falsey value, but still invalid props?

samanthavholmes: well it could be but there wasn't a check for invalid props before this change - i don't think the component would render if they were invalid

the point of this check was to determine if to show it at all

PR #541, opened 2017-12-13
Remove Static components, config, scripts

drewdrewthis:


kellywu:
whew lots of delete coded!

jmcolella:
@drewdrewthis what's the status here? Seems like we don't need this code in here anymore, would be great to get it out

PR #546, opened 2017-12-19
Background blocks and gray boxes

PepperTeasdale:
Since the gray box is a central part of the new branding, we made it
reusable here. A few caveats: the `GrayBox` is an extension of `Layout`,
and the `verticalPadding` and `leftOffset` variants use `StyledWrapper`
which can behave strangely with advanced usage. These features work best
when the props for offset/ padding are not going to change at runtime.

### DIFF
--- a/src/assets/stylesheets/utilities/_break_points.scss
+++ b/src/assets/stylesheets/utilities/_break_points.scss
@@ -42,3 +42,5 @@ $xxlarge-only: '#{$screen} and (min-width:#{lower-bound($xxlarge-range)}) and (m
 $small-and-medium-only: '#{$screen} and (max-width: #{upper-bound($tablet-range)})';
 $medium-up-to-large: '#{$screen} and (min-width:#{lower-bound($tablet-range)}) and (max-width:#{upper-bound($desktop-range)})';
 $medium-up-to-xxlarge: '#{$screen} and (min-width:#{lower-bound($tablet-range)}) and (max-width:#{upper-bound($xlarge-range)})';
+
+$IE: 'all and (-ms-high-contrast: none), (-ms-high-contrast: active)';
--- a/src/atoms/GrayBox/gray-box.module.scss
+++ b/src/atoms/GrayBox/gray-box.module.scss
@@ -1,0 +1,111 @@
+@mixin g-unit($cols: 1, $total-col-count: 12) {
+  $width: calc( ( #{$cols} / #{$total-col-count} ) * 100%);
+
+  width: $width;
+  flex-basis: $width;
+  max-width: $width;
+  flex-grow: 0;
+  flex-shrink: 1;
+}
+
+.gray-box {
+  position: relative;
+  padding: rem-calc(36) 0;
+
+  > div {
+    position: relative;
+  }
+
+  div:first-child {
+    @media #{$tablet} {
+      margin-left: calc(100% / 12);
+    }
+  }
+
+  &:after {
+    content: '';
+    position: absolute;
+    z-index: 1;
+    display: block;
+    height: rem-calc(110);
+    top: rem-calc(-70);
+    margin-bottom: rem-calc(-110);
+    right: calc(100% / 12);
+
+    @media #{$mobile-only} {
+      height: rem-calc(60);
+      top: rem-calc(-40);
+    }
+  }
+
+  &:before {
+    content: '';
+    position: absolute;
+    display: block;
+    top: 0;
+    bottom: 0;
+    background-color: color('neutral-3');
+    box-sizing: content-box;
+  }
+
+  @media #{$tablet} {
+    &.right {
+      div:first-child {
+        margin-left: auto;
+      }
+
+      &:before {
+        padding-right: #{$section-padding-desktop};
+        right: -#{$section-padding-desktop};
+      }
+
+      &:after {
+        right: 0;
+      }
+    }
+
+    &.left {
+      div:first-child {
+        margin-left: 0;
+      }
+
+      &:before {
+        left: -#{$section-padding-desktop};
+        padding-left: #{$section-padding-desktop};
+      }
+    }
+  }
+}
+
+@for $i from 1 through 12 {
+  .gray-box-lg-#{$i} {
+    &:before {
+      @include g-unit($i);
+    }
+
+    // Make color dash 1 column to the left of grey box right side
+    &:after {
+      right: calc(100% / 12 * (#{12 - $i + 1}));
+
+      @media #{$mobile-only} {
+        right: 100%;
+      }
+    }
+  }
+}
+
+@media #{$mobile-only} {
+  .gray-box {
+    &:before {
+      min-width: 100vw;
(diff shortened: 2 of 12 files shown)


src/assets/stylesheets/utilities/_break_points.scss
```diff
@@ -42,3 +42,17 @@ $xxlarge-only: '#{$screen} and (min-width:#{lower-bound($xxlarge-range)}) and (m
...
+// Use to compensate for scroll-to offset
+// This is the height of the main nav from mobile up
+$tablet-nav-height: 120px;
+
+$gutter-width: ru(.5);
+
+// -12px added to take into account gutter that comes with Layout in each section
+$section-padding-mobile: rem-calc(24px - 12px);
+$section-padding-tablet: rem-calc(48px - 12px);
+$section-padding-desktop: rem-calc(64px - 12px);
```
sunnymis: Can you do `rem-calc(24px - $gutter-width)` instead?

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
...
+.gray-box {
+  position: relative;
+  padding: rem-calc(36) 0;
+
+  > div {
+    position: relative;
+  }
+
+  div:first-child {
+    @media #{$tablet} {
```
sunnymis: This could be moved to the `@media #{$tablet}`  section on line 51

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
...
+    &:before {
+      @include g-unit($i);
+    }
+
+    // Make color dash 1 column to the left of grey box right side
+    &:after {
+      right: calc(100% / 12 * (#{12 - $i + 1}));
+
+      @media #{$mobile-only} {
+        right: 100%;
```
sunnymis: You can move this to the `@media #{$mobile-only }` section on line 97

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
...
+    position: absolute;
+    z-index: 1;
+    display: block;
+    height: rem-calc(110);
+    top: rem-calc(-70);
+    margin-bottom: rem-calc(-110);
+    right: calc(100% / 12);
+
+    @media #{$mobile-only} {
+      height: rem-calc(60);
```
sunnymis: You can move this to the @media #{$mobile-only } section on line 97

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
+@mixin g-unit($cols: 1, $total-col-count: 12) {
```
sunnymis: g-g-g-G-UNIT 🔥

src/atoms/GrayBox/index.js
```diff
@@ -0,0 +1,73 @@
...
+    >
+      <StyledWrapper css={styleOverrides} component={Col}>
+        {children}
+      </StyledWrapper>
+    </Layout>
+  );
+};
+
+GrayBox.propTypes = {
+  supertitle: PropTypes.string,
```
sunnymis: Add documentation to the prop types

src/assets/stylesheets/utilities/_break_points.scss
drewdrewthis: 👍 

Also, these can probably go into a variables file instead of the breakpoints. The `$IE` makes sense here, but, even though the others are related, they aren't really break points.

src/atoms/GrayBox/gray-box.module.scss
drewdrewthis: 🔥

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
...
+      right: calc(100% / 12 * (#{12 - $i + 1}));
+
+      @media #{$mobile-only} {
+        right: 100%;
+      }
+    }
+  }
+}
+
+@media #{$mobile-only} {
```
drewdrewthis: Consider moving the mobile-only section up above the other queries so that it's in a mobile first layout. Makes it less confusing to understand the cascade.

src/atoms/GrayBox/gray-box.module.scss
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  &:before {
+    content: '';
+    position: absolute;
+    display: block;
+    top: 0;
+    bottom: 0;
+    background-color: color('neutral-3');
+    box-sizing: content-box;
```
drewdrewthis: Yay! `content-box`!

src/atoms/GrayBox/index.js
```diff
@@ -0,0 +1,73 @@
...
+    if (leftOffset) {
+      str += `
+      @media ${breakpoints.mediumUp} {
+        margin-left: calc(100% / 12 * ${leftOffset}) !important;
+        max-width: calc(100% - (100% / 12 * ${cols - leftOffset}) !important;
+      }
+      `;
+    }
+
+    return str;
```
drewdrewthis: ☝️ This is slick.

src/atoms/GrayBox/index.js
```diff
@@ -0,0 +1,73 @@
...
+        padding-top: ${verticalPadding};
+        padding-bottom: ${verticalPadding};
+      `;
+    }
+
+    if (leftOffset) {
+      str += `
+      @media ${breakpoints.mediumUp} {
+        margin-left: calc(100% / 12 * ${leftOffset}) !important;
+        max-width: calc(100% - (100% / 12 * ${cols - leftOffset}) !important;
```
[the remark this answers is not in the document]
drewdrewthis: Discussed offline.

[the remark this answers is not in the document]
drewdrewthis: If you run the calculation, you’re actually just going to get a percentage. The only reason you use `calc()` is to calculate the arithmetic between two different units of measure in the browser. Otherwise, you do the math before you print to css

ie `calc(100% - 15px)` <-- ok vs `calc(100% - (100% / 12 * 15))` <-- not this

src/assets/stylesheets/utilities/_break_points.scss
PepperTeasdale: `rem-calc(24px - $gutter-width)` failed the build, i think because of mixing units. I also tried re-defining `$gutter-width` with `rem-calc`, same issue.

src/atoms/GrayBox/index.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+      `;
+    }
+
+    return str;
+  };
+
+  return (
+    <Layout
+      mediumCols={[ cols - 1 ]}
```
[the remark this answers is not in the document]
PepperTeasdale: Discussed offline. We can revisit later as we see how this is used in actual practice.

PR #549, opened 2017-12-21
Update block header styling and accept a description

samanthavholmes:


src/molecules/BlockHeader/index.js
```diff
@@ -2,59 +2,67 @@ import React from 'react';
...
 BlockHeader.propTypes = {
   /**
    * The small title above the main title
-  */
+   */
   supertitle: PropTypes.string,
   /**
-   * The main title
-  */
+   * the main title
```
sunnymis: I think this should be `The`

src/molecules/BlockHeader/index.js
```diff
@@ -2,59 +2,67 @@ import React from 'react';
...
+   */
   title: PropTypes.string,
+  /**
+   * the description under the title
```
sunnymis: Also `The`

PR #550, opened 2017-12-21
Augment functionality of IconList component

drewdrewthis:
We need the IconList to also be able to have a description. This allows
for that, while also making some css updates to allow the icons to align
correctly at different font-sizes.

### DIFF
--- a/src/atoms/Icon/index.js
+++ b/src/atoms/Icon/index.js
@@ -7,7 +7,11 @@ import * as icons from 'assets/images';
 import styles from './icons.module.scss';
 
 function Icon( props ) {
+
   const icon = icons[props.icon];
+
+  if (!icon) { return null; }
+
   const {
     className,
     onClick,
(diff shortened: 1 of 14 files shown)


src/molecules/StarRating/__tests__/__snapshots__/star_rating.spec.js.snap
```diff
@@ -3,56 +3,5 @@
...
-    onClick={undefined}
-    style={
-      Object {
-        "height": undefined,
-        "width": undefined,
-      }
-    }
-  />
-</span>
+/>
```
drewdrewthis: Yay! Cleaner code.

src/atoms/Icon/index.js
```diff
@@ -17,22 +21,22 @@ function Icon( props ) {
   } = props;
 
   return (
-    <span
+    <SVGInline
```
drewdrewthis: Yay! 1 less dom node!

src/molecules/lists/IconList/index.js
```diff
@@ -4,35 +4,53 @@ import classnames from 'classnames';
...
               {item.text}
+              {
+                item.tooltip ?
+                  <Tooltip
+                    className={styles.tooltip}
+                    left
+                    inline
+                  >
+                    {item.tooltip}
+                  </Tooltip> : ''
```
drewdrewthis: ☝️  I had to do this with a ternary instead of the more idiomatic `bool && component` structure because if you have a child and a `null` inside of a `Text` component, the app breaks and gets very sad. :(

drewdrewthis:
@n3m3sis42 >> LGTM!

PR #561, opened 2018-01-02
Update step indicator and footer for life-web

jmcolella:
@trevornelson CR please

[CH 13778](https://app.clubhouse.io/policygenius/story/13778/user-should-see-updated-headers-and-footers-throughout-the-life-flow)

- Updates `StepIndicator` to not center active step on mobile
- Updates `Footer` and adds conditional constraint to allow for flexibility when displaying Footer on larger screens

This PR does not include updates needed for the tablet Footer version. That is in a separate ticket and will be updated accordingly.

### DIFF
--- a/src/molecules/StepIndicator/index.js
+++ b/src/molecules/StepIndicator/index.js
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import get from 'lodash/get';
 import classnames from 'classnames';
-import Text from 'atoms/Text';
 import ProgressBarStep from './ProgressBarStep';
 import styles from './step_indicator.module.scss';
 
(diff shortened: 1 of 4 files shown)


src/molecules/StepIndicator/index.js
```diff
@@ -13,13 +11,8 @@ function StepIndicator(props) {
     navigateToPath
   } = props;
 
-  const activeStep = steps.find(step => step.currentStepActive) || steps[0];
-
   return (
     <div className={classnames(className)}>
-      <Text tag='div' type={11} color='primary-3' font='a' spaced className={styles['active-step']}>
```
trevornelson: Does the progress bar after the health navigator look OK still on iphone 5s if this is removed? That was why this was initially limited to only the active step. We may have shortened some of the breadcrumb titles though since so maybe it's not an issue anymore?

jmcolella: @trevornelson good point. Since we removed the `confirmation` step, the text works fine for iphone5. maybe this might be a limitation for the StepIndicator, that on mobile it can only accept up to 4 steps? Or we could put some control flow here to render this centered text if there are more than 4 steps? I think these are questions we can punt on for now since, to me, it seems pretty low priority

trevornelson: 👍  as long a it looks good with the current breadcrumbs we'll be using this should be all good.

PR #583, opened 2018-01-12
Update react peer dependency and yarn engine

jmcolella:
@drewdrewthis @samanthavholmes CR please

This updates the React `peerDependency` to include both versions 15 and 16 as well as allowing to use a yarn engine lower than `1.0.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     "whatwg-fetch": "^2.0.3"
   },
   "peerDependencies": {
-    "react": "^16.2.0"
+    "react": "^15.0.0 || ^16.0.0"
   },
   "engines": {
     "yarn": "^1.3.2",


package.json
```diff
@@ -154,10 +154,10 @@
     "whatwg-fetch": "^2.0.3"
   },
   "peerDependencies": {
-    "react": "^16.2.0"
+    "react": "^15.0.0 || ^16.0.0"
   },
   "engines": {
-    "yarn": "^1.3.2",
+    "yarn": "^0.24.0 || ^1.0.0",
```
drewdrewthis: @jmcolella We spoke about this offline. I think we should have locked-in engine versioning.

jmcolella:
Updated to lock in yarn engine to any patch within `1.3.x`

PR #586, opened 2018-01-17
Add GrayBox to Layout

sunnymis:
The current GrayBox component works great for it's intended purpose:
being a gray box that holds some content. However, sometimes we need a
gray box that is merely a background and doesn't care if there is
content inside. LayoutGrayBox is that. When placed inside a layout it
serves as a background to that layout and doesn't affect the other
columns within that layout

### DIFF
--- a/src/atoms/Layout/index.js
+++ b/src/atoms/Layout/index.js
@@ -163,3 +163,4 @@ Layout.defaultProps = {
 };
 
 export { Layout as default, Layout, Col };
+export { GreyBox } from './GreyBox';
--- a/src/atoms/Layout/layout.module.scss
+++ b/src/atoms/Layout/layout.module.scss
@@ -6,6 +6,18 @@ $spacer-base: .25;
     .col-#{$size}-#{$i} {
       @include g-unit($i);
     }
+
+    @media #{$small-only} {
+      .col-offset-#{$i} {
+        margin-left: 0;
+      }
+    }
+
+    @media #{$medium-up} {
+      .col-offset-#{$i} {
+        margin-left: calc((100% / 12) * #{$i});
+      }
+    }
   }
 
   .col-#{$size}-auto {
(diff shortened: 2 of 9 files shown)


src/atoms/Layout/LayoutGrayBox/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import styles from '../layout.module.scss';
+import Col from '../Col';
+
+function LayoutGrayBox(props) {
```
drewdrewthis: Consider just calling this `GrayBox`

src/atoms/Layout/LayoutGrayBox/index.js
```diff
@@ -0,0 +1,34 @@
...
+  } = props;
+
+  const classes = cx(
+    styles['gray-box'],
+    className,
+  );
+
+  return (
+    <Col
+      {...props}
```
drewdrewthis: Why not use `...rest` here?

src/atoms/Layout/index.js
```diff
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import { processChildren } from './utils';
 import styles from './layout.module.scss';
 import Col from './Col';
+import LayoutGrayBox from './LayoutGrayBox';
```
drewdrewthis: I think you can just do:

`export LayoutGrayBox from './LayoutGrayBox';` and not update the bottom.

src/atoms/Layout/layout.module.scss
```diff
@@ -67,6 +73,12 @@ $spacer-base: .25;
       }
     }
   }
+
+  @for $i from 1 through 12 {
```
drewdrewthis: Can you combine this loop with the one above, and put the `@media`s in here?

PR #589, opened 2018-01-17
Fix Tooltip Bug - Not Closing Modal on Mobile

sunnymis:
When you clicked the Tooltip on mobile and a modal popped up, users were
not able to close that modal. The Tooltips openModal function was firing
everytime a user clicked the X on the modal. This was because the modal
was nested inside a `<span>` with an onClick handler.

The solution was to move make the Modal a sibling of the Tooltip instead
of a child.


drewdrewthis:
LGTM

PR #592, opened 2018-01-23
WIP - Update js linter to meet Policygenius standards

Jexeones24:
Athenaeum's linter config was created before we came up with our
standards. This will bring it up to the new standards and make it easier
to use with other repos.


Jexeones24:
@drewdrewthis => We should remove as many of the rules that are specifically set in this package's config as possible, and allow the policygenius one to set the linter rules.

src/atoms/ErrorMessage/index.js
```diff
@@ -9,9 +9,9 @@ import Text from 'atoms/Text';
 import styles from './error_message.module.scss';
 
 const hocErrorMessage = BaseComponent => props =>
-  <TransitionGroup>
+  (<TransitionGroup>
```
jmcolella: @Jexeones24 is the linter forcing you to put the `(` on the same line as the jsx? I kinda hate how this looks and think it's silly...but what do i know. would much rather prefer:

```
const Component = (props) => (
  <Child {...props} />
);
```

drewdrewthis: We should follow the linter that's the PG standard. https://github.com/policygenius/athenaeum/pull/592 is the PR open to update Athenaeum to the new js linter

jmcolella: but who made this a standard? can we iterate on our standards? not sure why this garbage has to be a thing

PR #595, opened 2018-01-24
Update GrayBox component position

Jexeones24:
The amount we need to shift the `GrayBox`, left or right, depends on the
section padding. However section padding doesn't take into account the
12px gutter from `Layout`. So we have to add the 12px back

### DIFF
--- a/src/molecules/BlockHeader/GrayBox/gray-box.module.scss
+++ b/src/molecules/BlockHeader/GrayBox/gray-box.module.scss
@@ -59,8 +59,8 @@
       }
 
       &:before {
-        padding-right: #{$section-padding-desktop};
-        right: -#{$section-padding-desktop};
+        padding-right: #{$section-padding-desktop + $gutter-width};
+        right: -#{$section-padding-desktop + $gutter-width};
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/molecules/BlockHeader/GrayBox/gray-box.module.scss
```diff
@@ -59,8 +59,8 @@
       }
 
       &:before {
-        padding-right: #{$section-padding-desktop};
-        right: -#{$section-padding-desktop};
+        padding-right: #{$section-padding-desktop + $gutter-width};
+        right: -#{$section-padding-desktop + $gutter-width};
```
drewdrewthis: Consider extracting this out into a new variable. 
`$box-offset: $section-padding-desktop + $gutter-width`

src/molecules/BlockHeader/GrayBox/gray-box.module.scss
```diff
@@ -71,8 +71,8 @@
       }
 
       &:before {
-        left: -#{$section-padding-desktop};
-        padding-left: #{$section-padding-desktop};
+        left: -#{$section-padding-desktop + $gutter-width};
+        padding-left: #{$section-padding-desktop + $gutter-width};
```
drewdrewthis: Same as above.

src/molecules/BlockHeader/GrayBox/gray-box.module.scss
Jexeones24: Sounds good.

PR #599, opened 2018-02-06
Add Google autocomplete field

jmcolella:
@trevornelson @drewdrewthis @sunnymis CR please

[CH 16362](https://app.clubhouse.io/policygenius/story/16362/shoppers-should-be-able-to-use-google-maps-autocomplete-for-address)

Adds a `GoogleAutoCompleteField` to the RCL.

This field mimics the one already implemented in renters, but makes some optimizations to grab data from Google based on props and to conditionally include the Google Places API to make the field work. The reasoning for the latter is that the consumer should have the power over where/how the Google Places script gets loaded in to avoid multiple loads of the script/collisions.

### DIFF
--- a/src/molecules/formfields/GoogleAutoCompleteField/Readme.md
+++ b/src/molecules/formfields/GoogleAutoCompleteField/Readme.md
@@ -1,0 +1,12 @@
+GoogleAutoCompleteField Example:
+
+**You must provide either a `googlePlacesLink` or `googlePlacesAPIKey` prop for this example to work**
+
+```jsx
+  <GoogleAutoCompleteField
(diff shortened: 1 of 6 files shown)


drewdrewthis:
Why should this go in the RCL?

jmcolella:
@drewdrewthis 

This component is now being used in both Renters and Life. By putting it in the RCL, we can more easily a) leverage this functionality in our other products/future products and b) iterate on this component in the future. Basically, it seems that adding Google autocomplete functionality to our address forms is becoming a pattern, so it would be ideal to house this logic in one place.

PR #606, opened 2018-02-16
Remove Modal header with hideHeader prop

jmcolella:
@trevornelson @Jexeones24 CR please


jmcolella:
gunna merge since this is tiny thanks guys

PR #610, opened 2018-02-20
Remove SVGs and node_modules from export [ch16835]

drewdrewthis:
To reduce the size of the RCL, this commit prevents
node_modules/dependencies from being bundled during export (this should
have already been happening.) This commit also drops SVGs from the
package and includes a library react-inlinesvg to server calls to get
the svgs on page load, as well as introduces a webpack bundle analyzer.


webpack.config.publish.js
```diff
@@ -11,33 +13,19 @@ module.exports = baseConfig({
...
-      minimize: false,
-      mangle: false,
-      compress: {
-        warnings: false
-      },
-      sourceMap: true
+    new UglifyJsPlugin({
+      sourceMap: false,
+      uglifyOptions: {
+        minimize: true,
```
PepperTeasdale: FWIW I think the default value is to minimize, but it is fine to be explicit here

webpack.config.publish.js
```diff
@@ -11,33 +13,19 @@ module.exports = baseConfig({
...
-      root: 'ReactDOM',
-      commonjs2: 'react-dom',
-      commonjs: 'react-dom',
-      amd: 'react-dom',
-      umd: 'react-dom',
-    },
-    lodash: 'lodash',
-    classnames: 'classnames'
-  },
+  externals: [ nodeExternals() ],
```
PepperTeasdale: How's it work if the consumer of this library does not have these `externals`? Is there an obvious error or does yarn know to install as a dependency of this dependency?

src/atoms/Trackable/utils.js
```diff
@@ -1,7 +1,7 @@
 import kebabCase from 'lodash/kebabCase';
 import assign from 'lodash/assign';
 
-export function dataAttrName(key) {
+export function dataAttrName(key = 'undefined') {
```
PepperTeasdale: What's this all about?

src/atoms/Img/__tests__/img.spec.js
```diff
@@ -5,6 +5,16 @@ import renderer from 'react-test-renderer';
 import Img from 'atoms/Img';
 
 describe('<Img />', () => {
+  const warn = console.warn;
+
+  beforeAll(() => {
+    console.warn = jest.fn();
+  });
```
PepperTeasdale: What's the intention of this?

src/atoms/Icon/index.js
```diff
@@ -1,54 +1,84 @@
...
+    style: {
+      width,
+      height
+    },
+    ...rest
+  };
+
+  const sanitizedIcon = icon.replace(/[^a-zA-Z0-9_-]/, '');
+
+  const src = `https://storage.googleapis.com/pg-static-assets/svgs/${sanitizedIcon}.svg`;
```
PepperTeasdale: Would there be any benefit to using imgix here potentially?

src/atoms/Icon/icons.module.scss
```diff
@@ -5,16 +5,12 @@
...
-  /* stylelint-disable selector-pseudo-class-no-unknown */
-  // This class comes from our SVG dependency
-  :global {
-    .SVGInline-svg {
-      width: 100%;
-      height: 100%;
-      display: block;
-    }
+  .img,
+  svg {
```
PepperTeasdale: Nice!

jestSetup.js
```diff
@@ -0,0 +1,3 @@
+global.requestAnimationFrame = function (callback) {
+  setTimeout(callback, 0);
+};
```
drewdrewthis: Gets rid of an annoying warning.

package.json
```diff
@@ -12,25 +12,11 @@
...
-    "clean:footer": "rm -rf ./footer/build/* && rm -rf ./footer/static/*",
-    "build:nav:styleguide": "./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.styleguide.js --progress --colors --env.static=main_nav",
-    "build:nav:bundle": "NODE_ENV=production ./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.bundle.js --progress --colors --env.static=main_nav",
-    "build:nav:static": "NODE_ENV=production ./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.server.js --progress --colors --env.static=main_nav",
-    "build:nav:html": "node main_nav/index.js",
-    "build:footer:styleguide": "./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.styleguide.js --progress --colors --env.static=footer",
-    "build:footer:bundle": "NODE_ENV=production ./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.bundle.js --progress --colors --env.static=footer",
-    "build:footer:static": "NODE_ENV=production ./node_modules/webpack/bin/webpack.js --config ./webpack.config.static.server.js --progress --colors --env.static=footer",
-    "build:footer:html": "node footer/index.js",
```
drewdrewthis: All this footer/nav stuff should have already been removed.

webpack.config.publish.js
drewdrewthis: lol, @PepperTeasdale I'm pretty sure this isn't even a real prop - I left it there just in case it is, but I don't think it does anything.

webpack.config.publish.js
drewdrewthis: @PepperTeasdale The package.json + yarn should handle that. It just prevents them from being bundled into index.js

src/atoms/Icon/index.js
drewdrewthis: Yes -- potentially. But I don't think it's a big enough win.

src/atoms/Img/__tests__/img.spec.js
drewdrewthis: This was warning you that the lazyload module wasn't available -- which it wouldn't be in the tests because we load it at page load.

src/atoms/Trackable/utils.js
drewdrewthis: @PepperTeasdale if key is `undefined` then `toLowerCase()` errors out. This will just make a `data-undefined` and move on. Not ideal.

PepperTeasdale: yeah, could we make it just not add anything instead? we were looking at some elements in gatsby earlier and there was a lot of "undefined" in class names/ data properties in the dom.

src/molecules/asides/ImageAside/index.js
```diff
@@ -4,7 +4,7 @@ import classnames from 'classnames';
 
 import Icon from 'atoms/Icon';
 import Text from 'atoms/Text';
-import * as icons from 'assets/images';
+// import * as icons from 'assets/images';
```
sunnymis: You need this?

PR #618, opened 2018-02-27
Add new ComparisonTable component

Jexeones24:
[CH16740](https://app.clubhouse.io/policygenius/story/16740/engineer-should-see-a-compare-component)

### DIFF
--- a/src/organisms/tables/ComparisonTable/index.js
+++ b/src/organisms/tables/ComparisonTable/index.js
@@ -1,24 +1,17 @@
 import React from 'react';
-import { Col, Layout } from 'atoms/Layout';
+import cx from 'classnames';
+import PropTypes from 'prop-types';
+import {
+  Col,
+  Layout,
+} from 'atoms/Layout';
+import Text from 'atoms/Text';
+import Spacer from 'atoms/Spacer';
+import Tooltip from 'atoms/Tooltip';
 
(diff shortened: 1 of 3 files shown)


src/organisms/tables/ComparisonTable/index.js
```diff
@@ -1,35 +1,144 @@
...
+      <Col />
       { React.Children.map(children, renderCol) }
     </Layout>
   );
 };
 
-const TableHeader = props => renderRow(props, 'col-head');
-const TableRow = props => renderRow(props);
+const TableHeader = props => props.children.length === 2 ? renderHeaderRow(props, 'col-header-2-cells') : renderRow(props, 'col-header-3-cells');
+const TableRow = props => renderRow(props, 'cell');
```
trevornelson: Same thing with `renderRow`- you could probably just move `renderRow`'s function to `TableRow`, then in `TableHeader` call `TableRow` instead of `renderRow`?

src/organisms/tables/_ComparisonTable/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
trevornelson: Since `_ComparisonTable` isn't being used anywhere I think, we could probably just delete it's files.

PR #624, opened 2018-03-02
Fix comparison table stitch

jmcolella:
@trevornelson @danielnovograd @Jexeones24 @drewdrewthis CR please

- Fixes issue where comparison table header stitch would have a weird 'box' effect due to the use of `box-shadow`. The specific use of widths at each breakpoint is to ensure that the stitch mimics the `line-height` of the text it is supposed to emphasize.

[Weird box effect](https://cl.ly/0U180M3z1W2P)


jmcolella:
Going to merge this in, post-merge feedback welcome

PR #626, opened 2018-03-05
Fix Sections undefined background image

sunnymis:
If no backgroundImageUrl was supplied to Section, it was inlining a
background-image url('undefined')

This was also creating multiple errors in Sentry

### DIFF
--- a/src/atoms/Section/index.js
+++ b/src/atoms/Section/index.js
@@ -36,7 +36,7 @@ function Section(props) {
       className={classnames(...classes)}
       style={{
         ...style,
-        backgroundImage: `url(${backgroundImageUrl})`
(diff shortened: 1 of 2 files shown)


src/atoms/Section/index.js
```diff
@@ -36,7 +36,7 @@ function Section(props) {
       className={classnames(...classes)}
       style={{
         ...style,
-        backgroundImage: `url(${backgroundImageUrl})`
+        backgroundImage: backgroundImageUrl ? `url(${backgroundImageUrl})` : undefined
```
drewdrewthis: You could also do something like : 
```
if (backgroundImageUrl) { 
   style.backgroundImage = `url(${backgroundImageUrl})`
}
```

PR #627, opened 2018-03-12
Update ToggleField display logic [ch17366]

drewdrewthis:
For our Life4 EasyQuiz. We need to check that the input is actually
undefined, not that it's falsey


drewdrewthis:
@jmcolella << says that `==` is cool, yo.

PR #630, opened 2018-03-15
Modify ToggleField and Modal components

Jexeones24:
[CH16965](https://app.clubhouse.io/policygenius/story/16965/shopper-should-see-design-polish-updates-to-the-easy-quiz)

Adds a `nested` prop to remove ~6px left/right margin from Layout component wrapping the buttons and `sideBySide` prop to prevent the buttons from stacking on mobile. Also, added 60px of margin-top to x-large `Modal` variant as this is the one being used in the Easy Quiz.

### DIFF
--- a/src/molecules/formfields/ToggleField/index.js
+++ b/src/molecules/formfields/ToggleField/index.js
@@ -51,6 +51,8 @@ function ToggleField( props ) {
     input,
     meta,
     noBorder,
+    nested,
+    sideBySide,
   } = props;
 
   const classes = [
@@ -64,10 +66,14 @@ function ToggleField( props ) {
 
   const message = meta && meta.touched && !meta.active && (meta.error || meta.warning);
 
+  const buttonStyle = sideBySide ? styles['side-by-side'] : styles['button-wrapper'];
+
   return (
     <div>
       <div className={classnames(...classes)}>
-        <Layout>
+        <Layout
+          nested={nested}
+        >
           { label &&
             <Col className={styles['header']}>
               { tooltip &&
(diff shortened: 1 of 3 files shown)


src/molecules/formfields/ToggleField/index.js
```diff
@@ -64,10 +66,14 @@ function ToggleField( props ) {
...
   const message = meta && meta.touched && !meta.active && (meta.error || meta.warning);
 
+  const buttonStyle = sideBySide ? styles['side-by-side'] : styles['button-wrapper'];
+
   return (
     <div>
       <div className={classnames(...classes)}>
-        <Layout>
+        <Layout
+          nested={nested}
```
drewdrewthis: Should this just always be nested? Also, why are there so many divs here?

Jexeones24: Idk. But, was afraid to remove anything for fear of maybe breaking other things? I can change it if need be??

drewdrewthis: That's fair. I'd check with @jmcolella

jmcolella: no it shouldn't always be nested. when we have the border surrounding the toggle buttons we need the outside padding. The extra div is what provides the border to the buttons and then the top level div wraps the entire toggle field and the associated error message.

drewdrewthis: #boss man.

Jexeones24: Alright. Verdict is to leave it. Is this LGTM otherwise?

PR #633, opened 2018-03-23
Make react-styleguidist a devDependency

drewdrewthis:
React-styleguidist is only necessary for creating the RCL app, but it's
not necessary as part of the npm package we create.


PepperTeasdale:
LGTM

PR #634, opened 2018-03-28
Fix fragment and proptype errors

danielnovograd:
Cleans up some errors we were getting in life-web due to fragment array syntax and deprecated prop types.

[Corresponding PR](https://github.com/policygenius/life-web/pull/894)

@jmcolella @drewdrewthis @Jexeones24

### DIFF
--- a/src/organisms/cards/SimplePolicyCard/index.js
+++ b/src/organisms/cards/SimplePolicyCard/index.js
@@ -46,13 +46,11 @@ function SimplePolicyCard(props) {
         <div className={styles['body']}>
           <div className={styles['compare-logo-section']}>
             {
-              !isEmpty(compareCheckbox) ?
-              [
-                <Compare {...compareCheckbox} />,
+              !isEmpty(compareCheckbox) &&
+              <React.Fragment>
+                <Compare {...compareCheckbox} />
                 <div className={styles['divider']} />
-              ]
-              :
-              null
+              </React.Fragment>
             }
             <CarrierLogo carrierLogo={carrierLogo} />
           </div>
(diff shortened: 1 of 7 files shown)


src/organisms/cards/SimplePolicyCard/index.js
```diff
@@ -47,10 +47,10 @@ function SimplePolicyCard(props) {
           <div className={styles['compare-logo-section']}>
             {
               !isEmpty(compareCheckbox) ?
-              [
-                <Compare {...compareCheckbox} />,
-                <div className={styles['divider']} />
-              ]
+                <React.Fragment>
```
drewdrewthis: We should change this to the `&&` syntax

src/templates/Footer/index.js
```diff
@@ -82,8 +82,9 @@ function renderPhoneInfo(phoneNumber, hours) {
 
       <div className={styles['hours']}>
         {
-          hours && hours.map(hour =>
+          hours && hours.map((hour, i) =>
             <Text
+              key={`hours-${i}`}
```
drewdrewthis: It's bad practice, I believe, to use the map index as a key. Just use `hour`?

drewdrewthis: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

danielnovograd: It doesn't really matter for a component that will never get updated but OK FINE

drewdrewthis: haha

drewdrewthis: I'm surprised we don't have this linter rule in place.

PR #635, opened 2018-04-03
Update all stylesheet imports to relative

jmcolella:
@danielnovograd @drewdrewthis @trevornelson CR please

In order to properly import athenaeum stylesheets into other services, life `life-web`, all internal scss imports within the athenaeum stylesheets must use relative paths instead of absolute paths. The main issue is that, when trying to import a stylesheet that uses an absolute path, webpack cannot resolve that path correctly. It seems that other major UI libraries, like Bootstrap, have organized their internal scss stylesheets in this manner and have allowed consumers to import their stylesheets without issue.

I have tested this in both `life-web` and locally in `athenaeum`. This new relative path system should work. We will need to update `pg-gatsby`'s imports once this is merged in.


drewdrewthis:
@jmcolella You're a boss.

PR #644, opened 2018-05-04
Switch to GCS backend

PepperTeasdale:


bin/build
```diff
@@ -0,0 +1,23 @@
...
+  echo '+++ Publishing the package'
+  echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
+  npm whoami
+  BUMP_SIZE=$(buildkite-agent meta-data get bump-size)
+  VERSION=$(npm version $BUMP_SIZE)
+
+  bin/build_assets
+  npm publish --registry=https://registry.npmjs.org/
+else
+  echo "Not publishing to npm for branch $BUILDKITE_BRANCH"
```
drewdrewthis: lol. How informative.

bin/deploy
```diff
@@ -1,45 +1,23 @@
...
-echo '+++ Deploying styleguide latest to S3'
-aws s3 sync \
-  ./styleguide s3://rcl-pg/latest/ \
-  --acl=public-read \
-  --cache-control max-age=0 \
-  --exclude '.DS_Store'
+echo '+++ Deploying styleguide latest to GCS'
+gsutil -m -h "Cache-Control:public,max-age=0" \
+  cp -r -z html,css,js,gz -a public-read \
+  ./styleguide/. gs://rcl/latest/
```
drewdrewthis: I don't think we need this latest business, do we? Can't latest just be in the root?

PepperTeasdale: Probably not, but I want to get this in and working first, with minimal changes. I'll create a spinoff ticket to clean that part up a bit.

PR #647, opened 2018-05-16
Update Modal for improved experience

jmcolella:
@trevornelson @danielnovograd @drewdrewthis @Jexeones24 CR please

[CH 20133](https://app.clubhouse.io/policygenius/story/20133/shoppers-should-have-an-improved-modal-experience)

- Adds `overflow: hidden` to `body` when Modal opens to prevent background scrolling
- Updates header of Modal to remove grey background
- Updates `margin-top` for generic Modal


src/molecules/Modal/index.js
```diff
@@ -29,6 +29,26 @@ class Modal extends React.Component {
     }
   }
 
+  onAfterOpen = () => {
+    const { onAfterOpen } = this.props;
+
+    document.body.style.overflow = 'hidden';
+
+    if (onAfterOpen) {
+      onAfterOpen();
```
drewdrewthis: Also, since this is part of the ReactModal api which is already exposed, why do we need to force this on the user, instead of just allowing them to pass it in in themselves?

If you want this to be the default behavior, I suggest considering a defaultProp declaration.

src/molecules/Modal/index.js
```diff
@@ -38,34 +58,26 @@ class Modal extends React.Component {
...
-        onRequestClose={onRequestClose}
+        onAfterOpen={this.onAfterOpen}
+        onRequestClose={this.onRequestClose}
         contentLabel={contentLabel}
         className={classnames(styles['modal'], styles[variant], className)}
         overlayClassName={styles['overlay']}
+        shouldFocusAfterRender={false}
       >
-        <div className={styles['dialog']}>
+        <div id={contentLabel} className={styles['dialog']}>
```
drewdrewthis: Is there a chance that this won't be unique? Do we call this somewhere? Does it make more sense to expose and pass down an ID prop?

src/molecules/Modal/index.js
jmcolella: @drewdrewthis not sure what you're getting at here? I'm still allowing the user to pass this in or not. If you look below, I'm not passing the `onAfterOpen` prop to React Modal, I'm now passing in this instance method instead, so I'm ensuring that the user's `onAfterOpen` prop is still called if provided

src/molecules/Modal/util/mobileInputScroll.js
```diff
@@ -0,0 +1,23 @@
+import elementDistanceFromTop from 'utils/elementDistanceFromTop';
+
+export default (dialogId) => {
+  if (navigator.maxTouchPoints) {
+    const scrollToInput = () => {
```
drewdrewthis: Since you don't use `dialogId` in this function, I would pull this function out of the default function as a stand-alone function.

src/utils/elementDistanceFromTop.js
```diff
@@ -0,0 +1,11 @@
+export default (element) => {
```
drewdrewthis: Consider using a memo here (though there is probably no real need to optimize this).

src/molecules/Modal/index.js
```diff
@@ -23,12 +24,50 @@ const wrapChild = (child) => {
...
     }
   }
 
+  componentDidMount() {
+    const { enableInputScroll } = this.props;
+
+    if (enableInputScroll) {
+      const { dialogId } = this.props;
+
+      mobileInputScroll(dialogId);
```
drewdrewthis: Naming suggestion: `enableMobileScrollToInput`

PR #651, opened 2018-05-22
Fix modal background scroll for iOS devices

jmcolella:
@drewdrewthis @danielnovograd @Jexeones24 @trevornelson CR please

When updating the `Modal` to not allow for a background scroll, iOS devices did not respect the styling applied to the `body` to prevent that. This PR further updates the logic to prevent scroll by targeting iOS devices using the `navigator.userAgent` property and applying further `body` styling to ensure there is not background scroll when a modal is open on iOS. The logic also ensures that when a user closes the modal, they are brought back to the spot on the page where the modal had initially opened.

### DIFF
--- a/src/molecules/Modal/index.js
+++ b/src/molecules/Modal/index.js
@@ -8,6 +8,7 @@ import Icon from 'atoms/Icon';
 import MobileMenu from 'molecules/MobileMenu';
 import styles from './modal.module.scss';
 import mobileScrollToInput from './util/mobileScrollToInput';
+import checkiOSDevice from './util/checkiOSDevice';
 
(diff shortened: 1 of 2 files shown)


src/molecules/Modal/index.js
```diff
@@ -63,6 +73,11 @@ class Modal extends React.Component {
 
     document.body.style.overflow = this.state.initialBodyOverflow;
 
+    if (checkiOSDevice(navigator.userAgent)) {
```
Jexeones24: Cool. Had no idea what `navigator.userAgent` was, but just looked it up. Interesting!

PR #662, opened 2018-07-09
Add lazy flag for each item in IconList

jmcolella:
@danielnovograd @Jexeones24 @sunnymis CR please

Adds the ability to pass a `lazy` flag for each item in `data` supplied to the `IconList`. This gives us the flexibility to use the `IconList` when doing something like SSR, where lazy load is not necessary.

### DIFF
--- a/src/molecules/lists/IconList/index.js
+++ b/src/molecules/lists/IconList/index.js
@@ -35,6 +35,7 @@ function IconList(props) {
               <Icon
                 className={styles.icon}
                 icon={item.icon}
+                lazy={typeof item.lazy === 'undefined' ? true : item.lazy}
                 inline
               />
               {item.text}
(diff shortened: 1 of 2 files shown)


src/molecules/lists/IconList/index.js
```diff
@@ -35,6 +35,7 @@ function IconList(props) {
               <Icon
                 className={styles.icon}
                 icon={item.icon}
+                lazy={typeof item.lazy === 'undefined' ? true : item.lazy}
```
danielnovograd: What's your thought on just passing this as a defaultProp?

jmcolella: @danielnovograd I don't think you can pass in a `defaultProp` for a specific value in an object, the object here being the `data` prop.

PR #664, opened 2018-07-11
Add optional subtitles to StepIndicator

danielnovograd:
- Adds `inert` prop to the StepIndicator to allow the creation of presentational, non-interactive step indicators.
- Allows for the addition of subtitles for each step

Corresponds to [CH21562](https://app.clubhouse.io/policygenius/story/21562/user-should-see-a-slimmed-full-app-including-exit-survey)

@jmcolella @Jexeones24

### DIFF
--- a/src/molecules/StepIndicator/ProgressBarStep.js
+++ b/src/molecules/StepIndicator/ProgressBarStep.js
@@ -5,21 +5,25 @@ import classnames from 'classnames';
 import styles from './step_indicator.module.scss';
 
 export default class ProgressBarStep extends Component {
-  textProps() {
+  titleProps() {
     const {
-      step
+      step: { currentStepActive }
     } = this.props;
 
-    if (step.currentStepActive) {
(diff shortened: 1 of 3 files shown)


src/molecules/StepIndicator/ProgressBarStep.js
```diff
@@ -49,44 +53,57 @@ export default class ProgressBarStep extends Component {
 
   textClasses() {
     const {
+      inert,
```
jmcolella: could we not haha. `unclickable` probably better. maybe even `static`

src/molecules/StepIndicator/step_indicator.module.scss
```diff
@@ -85,6 +85,12 @@
   text-transform: uppercase;
 }
 
+.step-subtitle {
+  margin-top: 2rem;
```
jmcolella: `rem-calc`

PR #667, opened 2018-07-19
make numbered accordion list optional

cisacke:
one requirement of [this Accordion ticket](https://app.clubhouse.io/policygenius/story/21829/content-editors-can-create-accordion-sections) is that the numbers should not be hardcoded, so I've added in a prop to disable the numbered lists.

👀  please @sunnymis @jl08

### DIFF
--- a/src/organisms/Accordion/index.js
+++ b/src/organisms/Accordion/index.js
@@ -4,7 +4,7 @@ import cx from 'classnames';
 import styles from './accordion.module.scss';
 import AccordionPanel from './AccordionPanel';
 
-const sections = ({ panels, openText, closeText }) =>
+const sections = ({ panels, openText, closeText, disableNumberedList }) =>
   panels.map((panel, i) =>
     <AccordionPanel
       openText={openText}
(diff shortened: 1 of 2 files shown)


src/organisms/Accordion/index.js
```diff
@@ -67,12 +70,17 @@ Accordion.propTypes = {
...
+   * Whether or not to number the accordion items. Defaults to `true`.
+   */
+  numberedList: PropTypes.bool
 };
 
 Accordion.defaultProps = {
   openText: 'View',
-  closeText: 'Close'
+  closeText: 'Close',
+  numberedList: true
```
sunnymis: Since this defaults to true we can assume that the user wouldn't need to specify it in props. Would it make more sense to rename this field `disableNumberedList` 
i.e `<Accordion disableNumberedList />`

cisacke: Yeah I was kinda on the fence about which way around it should. I think that makes sense tho, I'll switch it around 👍

PR #668, opened 2018-07-25
Update tooltip functionality

Jexeones24:
[CH22026](https://app.clubhouse.io/policygenius/story/22026/chore-implement-athenaeum-tooltip-in-life-web)

- Borrows functionality used in LifeWeb `Tooltip` to reveal `hover message` contents onClick when `revealOnClick` prop is passed to tooltip, otherwise default behavior is to show message on hover when screen size is not mobile 

- Sets a reference to the node using `setWrapperRef`. Ref is then used to detect an outside click to close the tooltip.

- Gives `DateField` component ability to render a `Tooltip`

- Ran into a problem with `CheckBoxField` where upon selection, the checkmark was not showing. Refactored the label text from a `Text` component using `tag='span'` to a wrapper span with the appropriate classname and functionality to toggle text color and font-weight.

CR @danielnovograd @jmcolella @jdanz

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "athenaeum",
-  "version": "8.14.1",
+  "version": "8.14.2",
   "description": "PolicyGenius React Component Library (RCL)",
   "main": "lib/index.js",
   "scripts": {
--- a/src/atoms/Tooltip/__tests__/tooltip.spec.js
+++ b/src/atoms/Tooltip/__tests__/tooltip.spec.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import { mount, shallow } from 'enzyme';
+import Icon from 'atoms/Icon';
+import Modal from 'molecules/Modal';
+import Tooltip from '../';
+
+describe('Tooltip', () => {
+  let component;
+  let props;
+
+  it('defaults to Icon when text is not provided', () => {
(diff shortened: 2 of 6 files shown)


src/molecules/formfields/CheckBoxField/index.js
```diff
@@ -27,14 +27,15 @@ class CheckBoxField extends React.Component {
...
-        <Text
-          tag='span'
-          bold={input.checked}
-          color={input.checked ? 'primary-3' : textColor || 'neutral-2'}
-          className={styles['checkbox-label']}
-        >
-          { label }
-        </Text>
+
+        <span className={styles['checkbox-label']}>
```
jmcolella: I would not change this please, I think it will break feature specs. The `tag` prop on `Text` should render a `span` for you. Double check, because it's working for me

src/molecules/formfields/DateField/date_field.module.scss
```diff
@@ -13,23 +13,23 @@ $border: 1px solid color('neutral-4');
 }
 
 .label-wrapper {
+  display: block;
```
jmcolella: I'm a little confused why this was updated, it seems to break the field when a `subLabel` is added. Can you rebase from some of the updates I made when putting all Athenaeum form fields into life-web? That might clean things up for this PR

PR #674, opened 2018-08-06
Fix width of inputs in multi-field components for IE

jmcolella:
@danielnovograd @Jexeones24 @jdanz CR please

Necessary input width updates for multi-field components to fix bugs in [CH 19023](https://app.clubhouse.io/policygenius/story/19023/chore-implement-modularized-styles-for-navigator)

### DIFF
--- a/src/molecules/formfields/DateField/index.js
+++ b/src/molecules/formfields/DateField/index.js
@@ -22,13 +22,22 @@ class DateField extends React.Component {
       tooltip,
     } = this.props;
 
+    const numberOfChildren = React.Children.toArray(children).length;
+
     const wrapChild = (child) => {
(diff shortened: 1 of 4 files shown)


src/molecules/formfields/DateField/index.js
```diff
@@ -22,13 +22,22 @@ class DateField extends React.Component {
       tooltip,
     } = this.props;
 
+    const numberOfChildren = React.Children.toArray(children).length;
+
     const wrapChild = (child) => {
       const classes = [
         styles['input'],
+        styles[`fields-${numberOfChildren}`],
```
Jexeones24: I like this, very clever!

PR #677, opened 2018-08-09
Add renderProp to RadioCard component

jmcolella:
@danielnovograd @Jexeones24 @jdanz CR please

Relates to [CH 22177](https://app.clubhouse.io/policygenius/story/22177/user-should-be-prompted-with-a-page-to-fork-coverage-experience)

Adds `renderProp` prop to `RadioCard` to allow for more flexible styling of UI. The `renderProp` takes a function that must return a React node and currently passes to that function as arguments a) `radioButton`, which is the actual UI/functionality of the circular button,  b) `selected`, which is a boolean of whether the card is selected and c) `notSelected`, which is a boolean opposite of selected.

Also adds 2 new icons to complete the design in the ticket.


src/organisms/cards/RadioCard/index.js
```diff
@@ -132,7 +165,15 @@ RadioCard.propTypes = {
...
-  label: PropTypes.string.isRequired,
+  label: (props) => {
+    if (!props['label'] && !props['renderProp']) {
+      return new Error(
+        'Label prop is required',
+      );
+    }
+
+    return null;
+  },
```
Jexeones24: Woah. This is pretty cool - a custom propType?

jmcolella: yeah, added it because `label` should still be required but only when `renderProp` is not defined in the props

PR #700, opened 2018-10-09
Add DateField render prop and parsing

jmcolella:
@sunnymis @trevornelson CR please

Adds a `render` prop to the `DateField` component as a means to provide the necessary functionality to handle all date string parsing for our typical date field elements. The component handles parsing a UTC Date string into separate day, month and year values. The `render` prop exposes input props for each field that hook into the components internal logic for handling change and error states for the `DateField` as a whole. This will work for both full and short date fields.


jmcolella:
CORRECTION:

This PR creates 2 new DateField components: `FullDateField` and `ShortDateField`. Each field takes a `render` prop that exposes the individual fields (with full functionality baked in) to build any date field. I split up the the fields into 2 separate components because the update/onChange logic was different enough to make separate.

src/molecules/formfields/FullDateField/full_date_field.module.scss
```diff
@@ -0,0 +1,3 @@
+.full-date-field {
```
trevornelson: Do we need this file if there's no styles here?

src/molecules/formfields/ShortDateField/short_date_field.module.scss
```diff
@@ -0,0 +1,3 @@
+.short-date-field {
+  // Place your component specific scss here.
```
trevornelson: same here

src/molecules/formfields/FullDateField/full_date_field.module.scss
jmcolella: nope!

PR #702, opened 2018-10-29
Bug - Wrong Date on FullDateField when 1st of Month

sunnymis:
The `FullDateField` was displaying the previous day when passed a date that was the first of the month. This has to do with how `new Date()`. works, it does some auto-adjusting behind the scenes. It is safer to pass the date values directly into the constructor i.e `new Date(2018, 10, 29)`

### DIFF
--- a/src/molecules/formfields/util/BaseDateField/index.js
+++ b/src/molecules/formfields/util/BaseDateField/index.js
@@ -21,7 +21,7 @@ class BaseDateField extends React.Component {
   constructor(props) {
     super(props);
 
-    const date = new Date(props.input.value);
+    const date = this.createDate(props.input.value);
 
     if (isValid(date)) {
       this.state = {
@@ -36,6 +36,17 @@ class BaseDateField extends React.Component {
         yearValue: ''
       };
     }
+  }
+
+  createDate = (date) => {
+    /* New dates are in UTC time and default to midnight if not specified.
+     * EST is behind UTC so it'll calculate the date as the previous date.
+     * We need to offset that time correctly assign the date.
+     * See https://stackoverflow.com/questions/45407072/javascript-dates-are-a-day-off
+     */
+    const utcDate = new Date(date);
+
+    return new Date(utcDate.getTime() + (utcDate.getTimezoneOffset() * 60000));
   }
 
   changeFullDate = (prevState) => {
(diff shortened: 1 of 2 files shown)


src/molecules/formfields/util/BaseDateField/index.js
```diff
@@ -38,6 +38,19 @@ class BaseDateField extends React.Component {
...
+  createDate = (date) => {
+    const parsedDate = date.split('-');
+    if (parsedDate.length === 3) {
+      const year = parsedDate[0];
+      const month = parsedDate[1];
+      // Split day with timestamp if it is in ISO-8601 format
+      const day = parsedDate[2].split('T')[0];
+      return new Date(year, month - 1, day);
+    } else {
+      return new Date(date);
```
sunnymis: If for whatever reason we can't parse this date, i'm just passing it directly into `new Date()` and let it handle it. Doubt it'd ever happen though.

src/molecules/formfields/util/BaseDateField/index.js
```diff
@@ -38,6 +38,21 @@ class BaseDateField extends React.Component {
     }
   }
 
+  createDate = (date) => {
+    if (!date) { return new Date(date); }
+
+    let parsedDate = date.split('-');
```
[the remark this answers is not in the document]
sunnymis: @trevornelson `new Date()` handled all the parsing for us. Not sure how we'd get the three numbers if we don't parse a string. But I'll see if theres another way though

src/molecules/formfields/util/BaseDateField/index.js
```diff
@@ -38,6 +38,18 @@ class BaseDateField extends React.Component {
...
+  createDate = (date) => {
+    /* New dates are in UTC time and default to midnight if not specified.
+     * EST is behind UTC so it'll calculate the date as the previous date.
+     * We need to offset that time correctly assign the date.
+     * See https://stackoverflow.com/questions/45407072/javascript-dates-are-a-day-off
+     */
+    const utcDate = new Date(date);
+
+
+    return new Date(utcDate.getTime() + (utcDate.getTimezoneOffset() * 60000));
```
jmcolella: ayyyy noice 👍

trevornelson: 👍

PR #703, opened 2018-11-01
Bug Fix - Prevent 0 Month or Day on FullDateField

sunnymis:
Prevent the user from entering 0 for the month or day in the `FullDateField`


jmcolella:
@sunnymis LGTM

PR #775, opened 2019-07-24
Add new icons for Coverage Needs

gummoe:
Story: https://app.clubhouse.io/policygenius/story/33850/60-users-should-see-a-needs-calculation-experience

Adds new entries into the constants file for Athenaeum icons.


jl08:
@gummoe too radical

PR #778, opened 2019-08-02
re-add Checkbox field and checkbox wrapper

jl08:
**Purpose**
Need to use `CheckboxWrapper` and `CheckBoxField` components for a feature in `pg-gatsby` 
Instead of copying this code over to use in `pg-gatsby` only, thought it would be better to just re-add it to this slimmed down version of `athenaeum`

**Fix**
- Add the `CheckboxWrapper` and `CheckBoxField` ( why is the casing inconsistent.. -.- ) that exists on master to this branch

@gummoe @jdanz @maxnovak @antciccone

### DIFF
--- a/src/molecules/formfields/CheckBoxField/Readme.md
+++ b/src/molecules/formfields/CheckBoxField/Readme.md
@@ -1,0 +1,10 @@
+Checkbox Example:
+
+```jsx
+  <CheckBoxField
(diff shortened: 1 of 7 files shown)


jl08:
> @jl08 - is this simply a copy paste job?

yes

gummoe:
Excellent

PR #798, opened 2019-10-10
NPM Saga: Part 3 

magcurt8:
Yeah so we had `npm` in our build steps for publishing to npm, removed those in favor of yarn since we are a `yarn shop now dog` - Ben Gummoe


magcurt8:
despite Ben's glorious comment, this pr is no longer relevant

PR #803, opened 2019-11-13
WINNE-129: Handling "invalid" street addresses

kengorab:
- There are some street addresses that google doesn't recognize, and therefore does not return the street_number for
    - See this for more info: https://stackoverflow.com/a/48545776
  - In these cases, attempt to extract the street number from the user-input value, assuming the street number will be the number at the start of their input. If such an assumption is invalid, return no street number

### DIFF
--- a/src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
+++ b/src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
@@ -1,16 +1,17 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import renderer from 'react-test-renderer';
+import { omit } from 'lodash';
 import TextField from 'molecules/formfields/TextField';
 
 import GoogleAutoCompleteField from '../';
 
 describe('<GoogleAutoCompleteField />', () => {
-  let props;
+  let baseProps;
   let addressValues;
 
   beforeEach(() => {
-    props = {
+    baseProps = {
       input: {
         name: 'streetAddress',
       },
@@ -59,7 +60,7 @@ describe('<GoogleAutoCompleteField />', () => {
   });
 
   it('renders correctly', () => {
-    const wrapper = shallow(<GoogleAutoCompleteField {...props} />);
+    const wrapper = shallow(<GoogleAutoCompleteField {...baseProps} />);
 
     const actual = renderer.create(wrapper).toJSON();
 
@@ -67,27 +68,27 @@ describe('<GoogleAutoCompleteField />', () => {
   });
 
   it('returns a single TextField with appropriate props', () => {
-    const wrapper = shallow(<GoogleAutoCompleteField {...props} />);
+    const wrapper = shallow(<GoogleAutoCompleteField {...baseProps} />);
 
     const field = wrapper.find(TextField);
 
     expect(field.length).toEqual(1);
-    expect(field.props().name).toEqual(props.name);
+    expect(field.props().name).toEqual(baseProps.name);
     expect(field.props().label).toEqual('Street address');
     expect(field.props().placeholder).toEqual('Enter your street address');
     expect(field.props().id).toEqual('autocomplete');
   });
 
   describe('updateAddress', () => {
-    it('calls autocomplete.getPlace and props.autocompleteAddressFields', () => {
-      const expectedData = {
-        number: '1',
-        street: 'Broadway',
-        city: 'New York',
-        state: 'New York',
-        zip: '10075',
-      };
+    const expectedData = {
+      number: '1',
+      street: 'Broadway',
+      city: 'New York',
+      state: 'New York',
+      zip: '10075',
+    };
 
+    it('calls autocomplete.getPlace and props.autocompleteAddressFields', () => {
       const state = {
         autocomplete: {
           getPlace: jest.fn(() => ({
@@ -96,6 +97,7 @@ describe('<GoogleAutoCompleteField />', () => {
         }
       };
 
+      const props = { ...baseProps };
       const wrapper = shallow(<GoogleAutoCompleteField {...props} />);
 
       wrapper.setState(state);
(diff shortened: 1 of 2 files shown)


src/molecules/formfields/GoogleAutoCompleteField/index.js
```diff
@@ -51,6 +51,21 @@ class GoogleAutoCompleteField extends Component {
   updateAddress = () => {
     const place = this.state.autocomplete.getPlace();
     const addressData = this.formatAddressInfo(place.address_components);
+
+    if (!addressData.number) {
```
kengorab: This is admittedly a little brittle, since it assumes that the user is providing valid input. Basically, if the Places API fails to return a `street_number` field (for background info, see [this link](https://stackoverflow.com/a/48545776)), we attempt to parse it ourselves from the user's input. To do this, we try to parse a block of digits at the start of their input. If this fails, then we don't return a street number, and we're no worse off.

src/molecules/formfields/GoogleAutoCompleteField/index.js
```diff
@@ -51,6 +51,21 @@ class GoogleAutoCompleteField extends Component {
...
+      const { value } = this.props.input;
+
+      if (value) {
+        const matches = value.match(/^(\d+)/);
```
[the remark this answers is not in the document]
parky22: Also could something like `123 west 50th street` return `12350` instead?

src/molecules/formfields/GoogleAutoCompleteField/index.js
```diff
@@ -51,6 +51,21 @@ class GoogleAutoCompleteField extends Component {
...
+
+        if (matches && matches[1]) {
+          const number = parseInt(matches[1], 10);
```
parky22: Is there a preference for returning something at index 1 instead of index 0? Also, is there an advantage of using `parseInt` over `Number()`?

src/molecules/formfields/GoogleAutoCompleteField/index.js
```diff
@@ -51,6 +51,21 @@ class GoogleAutoCompleteField extends Component {
...
+
+          addressData.number = { shortName: number, longName: number };
```
parky22: what's the difference between shortname and longname?

parky22: oh nevermind - just realized it's a Google thing

src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
```diff
@@ -106,6 +106,53 @@ describe('<GoogleAutoCompleteField />', () => {
...
+        const input = { name: 'address', value: 'Twelve Broadway' };
+        const wrapper = shallow(<GoogleAutoCompleteField {...props} {...{ id, input }} />);
+
+        wrapper.setState(state);
+        wrapper.instance().updateAddress();
+
+        expect(state.autocomplete.getPlace.mock.calls.length).toBe(1);
+        expect(props.autocompleteAddressFields.mock.calls.length).toBe(1);
+        expect(props.autocompleteAddressFields.mock.calls[0][0])
+          .toEqual({ ...expectedData, number: undefined });
```
parky22: Actually should we explicitly return `null` instead?

src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
```diff
@@ -106,6 +106,53 @@ describe('<GoogleAutoCompleteField />', () => {
...
+            }))
+          }
+        };
+      });
+
+      it('should extract the street_number manually', () => {
+        const streetNumber = 1234;
+        const id = 'autocomplete-id';
+        const input = { name: 'address', value: `${streetNumber} Broadw` };
+        const wrapper = shallow(<GoogleAutoCompleteField {...props} {...{ id, input }} />);
```
parky22: Just a suggestion, but another way that I've updated the props is something like: 
```
const testProps = {
   ...props,
   id,
   input,
};
<Component {...testProps} />
```

src/molecules/formfields/GoogleAutoCompleteField/index.js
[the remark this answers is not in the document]
kengorab: I'll make that first change, I like it! And no it shouldn't, because the regex will only capture sequential digits, when it hits the whitespace before `west` it'll stop. But I'll add a test for it just in case - it's good to have coverage there should anyone refactor this in the future!

src/molecules/formfields/GoogleAutoCompleteField/index.js
kengorab: Yeah, I keep it in this `{ shortName, longName }` format because further processing happens later on.

src/molecules/formfields/GoogleAutoCompleteField/index.js
kengorab: I actually never use `Number()` because I prefer not to instantiate classes of primitives using constructors; I've actually always just used `parseInt`/`parseFloat` in situations like this.

As for the first question, there actually isn't! Good catch, I'll change it - I had done `[1]` just out of habit to capture the first group, but since the first group is all that we're matching anyway, it's pretty redundant.

src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
kengorab: Yeah that looks a little more standard, I'll do that!

src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
kengorab: I'm not sure... I left it as `undefined` because that's how the original implementation worked if a street number were missing, and I wasn't sure if downstream code made any assumptions about non-present values. It's _probably_ safe though...

parky22: Gotcha! I personally like explicitly returning `null` instead of depending on the implicit return value. I think it sends a stronger signal that we are expecting this falsey value. An `undefined` can be a bit vague on whether the falsey value is intentionally being sent down or is coming unexpectedly.

kengorab: Lemme see if it causes anything to misbehave

kengorab: Well actually, I think omitting the `number` value _is_ the correct approach, because in this component you can pass in the fields you want to receive off of the google results, and if the street number isn't present there then it should be missing (aka `undefined`) rather than present but equal to `null`. In this test I was using `{ ...expectedData, number: undefined }` as a shorthand for
```js
omit(expectedData, 'number')
```
I'll update the test to make it more clear.

parky22: oohh i see - thanks for the clarification!

src/molecules/formfields/GoogleAutoCompleteField/__tests__/index.spec.js
```diff
@@ -106,6 +108,71 @@ describe('<GoogleAutoCompleteField />', () => {
...
+        expect(state.autocomplete.getPlace.mock.calls.length).toBe(1);
+        expect(props.autocompleteAddressFields.mock.calls.length).toBe(1);
+        expect(props.autocompleteAddressFields.mock.calls[0][0])
+          .toEqual({ ...expectedData, number: streetNumber });
+      });
+
+      it('should only extract the first digits as the street_number', () => {
+        const streetNumber = 12;
+        const id = 'autocomplete-id';
+        const input = { name: 'address', value: `${streetNumber} W 50th St` };
```
magcurt8: @sunnymis COLE HAAN

sunnymis: wut?

src/molecules/formfields/GoogleAutoCompleteField/index.js
```diff
@@ -51,6 +51,18 @@ class GoogleAutoCompleteField extends Component {
   updateAddress = () => {
     const place = this.state.autocomplete.getPlace();
     const addressData = this.formatAddressInfo(place.address_components);
+
+    if (!addressData.number) {
+      const { value } = this.props.input;
+      const matches = value && value.match(/^(\d+)/);
```
sunnymis: what are your thoughts on renaming this to something like `streetNumberMatch` just so it's clear what's being regex'ed for

kengorab: Yeah I like that, I'll make that change

PR #804, opened 2019-11-15
Norton should not be on our footer anymore

maxnovak:
We don't use Norton for security anymore and we should not be showing incorrect information in our footer.  We had removed this from our footer in all of our content, but our flows still show it.


maxnovak:
@gummoe It looks like it is still used in the `Checkout` template, which I have no idea where it is used.  Rather than possibly make a style somewhere look bad, going to keep it in for now.  But will see if I can figure out where that lives than will kill the whole thing.

PR #825, opened 2020-03-13
Fix bin/sonar to use /bin/sh instead of /bin/bash

randy-concepcion:
Small change to make `bin/sonar` script work. The sonar image needs to use `sh` instead of `bash`. This should fix the broken build.
 :man_facepalming:


randy-concepcion:
> shhhh - this never happened

:hushed: :speak_no_evil:

PR #829, opened 2020-03-31
WINNE-575: Adding `heldHammer` icon

kengorab:
- Also some minor cleanup for the styleguidist page. It removes the
`alert`s that pop up on page load (since tooltips, when passed as
functions, evaluate immediately).
- This also includes a slightly larger change, which is more of a
paradigm shift than the tooltip fix - this configures styleguidist to
display each component as a separate page rather than having one giant
page with all components. This breaks routing for old links, but I think
it's worth it because the page loads so much more quickly.

Here's the icon:
image

### DIFF
--- a/src/atoms/Icon/constants.js
+++ b/src/atoms/Icon/constants.js
@@ -129,6 +129,7 @@ module.exports = {
     'healthgenius',
     'healthgenius',
     'heart',
+    'heldHammer',
     'hmoPpo',
     'home',
     'homeAutoBundleColor',
--- a/src/molecules/formfields/CreditCardField/Readme.md
+++ b/src/molecules/formfields/CreditCardField/Readme.md
@@ -24,7 +24,7 @@ CreditCardField Example:
       securityCode={(
         <TextField
           placeholder='3/4 Digit Security Code'
-          inputTooltip={() => alert('input tooltip clicked')}
+          inputTooltip={'input tooltip clicked'}
           noBaseStyle
           input={{}}
         />
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -28,5 +28,6 @@ module.exports = {
     'holderjs',
     'assets/stylesheets/base.scss',
     path.join(__dirname, 'styleguide_assets/rcl_styles.module.scss')
-  ]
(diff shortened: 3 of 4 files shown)


src/atoms/Icon/constants.js
```diff
@@ -129,6 +129,7 @@ module.exports = {
     'healthgenius',
     'healthgenius',
     'heart',
+    'heldHammer',
```
kengorab: This is the main change, adding this icon (which is also included in the description).

src/molecules/formfields/CreditCardField/Readme.md
```diff
@@ -24,7 +24,7 @@ CreditCardField Example:
       securityCode={(
         <TextField
           placeholder='3/4 Digit Security Code'
-          inputTooltip={() => alert('input tooltip clicked')}
+          inputTooltip={'input tooltip clicked'}
```
kengorab: These changes remove the annoying popups that appear on page load

styleguide.config.js
```diff
@@ -28,5 +28,6 @@ module.exports = {
     'holderjs',
     'assets/stylesheets/base.scss',
     path.join(__dirname, 'styleguide_assets/rcl_styles.module.scss')
-  ]
+  ],
+  pagePerSection: true,
```
kengorab: This is a pretty nice quality-of-life change, it makes the page load _soo_ much faster, because each component's docs are on its own page rather than being concatenated into one humongous page that takes a while to load. Check out [the docs](https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#pagepersection)

src/molecules/formfields/CreditCardField/Readme.md
gummoe: you are a hero!

styleguide.config.js
gummoe: I really like this change! You might want to drop a quick note to the design team about this change when it gets merged in as well, only because they sometimes reference the RCL.

jmcolella: great change here, thanks!!!

PR #830, opened 2020-04-01
WINNE-575: Fixing button styles

kengorab:
- When a button of any `variant` is disabled, it should appear grayed
out. Previously, this styling had not been applied outlined,
variant=info buttons.
- Also, a button's icon (if present) should be colored to match the text
of the button.
- This also adjusts the styleguidist docs page for buttons, showcasing
the outlined and disabled forms of each variant.

### DIFF
--- a/src/atoms/Button/Readme.md
+++ b/src/atoms/Button/Readme.md
@@ -1,36 +1,102 @@
-Default Button Example:
+## Button Variants
+Buttons come in a few variants, which change their look-and-feel.
+The Buttons in these examples also have icons to demonstrate how the icon adapts to the variant.
 
+### Default
 ```jsx
-  <Button onClick={() => alert('Button clicked')} icon='lock'>Default Button</Button>
+<Button
+  icon='lock'
+  onClick={() => alert('Button clicked')}
+>
+  Default Button
+</Button>
 ```
 
-Info Button Example:
-
+#### Default Outlined
 ```jsx
-      <Button variant='info'>Info</Button>
+<Button
+  outline
+  icon='lock'
+  onClick={() => alert('Button clicked')}
+>
+  Outlined Default Button
+</Button>
 ```
 
-Outline Button Example:
-
+#### Default Disabled
 ```jsx
-      <Button outline>Outline Button</Button>
+<Button
+  disabled
+  icon='lock'
+  onClick={() => alert('Button clicked')}
+>
+  Disabled Default Button
+</Button>
 ```
 
-Outline Info Button Example:
-
+#### Default Outlined + Disabled
 ```jsx
-      <Button variant='info' outline>Info</Button>
+<Button
+  outline
+  disabled
+  icon='lock'
+  onClick={() => alert('Button clicked')}
+>
+  Disabled Outlined Default Button
+</Button>
 ```
 
-Disabled Unflexed Button Example:
-
+### Info
 ```jsx
-      <Button disabled unflex>Disabled</Button>
+<Button
+  variant='info'
+  icon='lock'
+  onClick={() => alert('Button clicked')}
+>
+  Info Button
+</Button>
 ```
 
-Disabled Unflexed Outline Button Example:
-
+#### Info Outlined
 ```jsx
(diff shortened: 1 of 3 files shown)


src/atoms/Button/Readme.md
```diff
@@ -1,36 +1,102 @@
-Default Button Example:
+## Button Variants
```
kengorab, line 1: I changed this page a bit to showcase all variants of the buttons. It looks like this now:
image

(Note this is zoomed out super far to fit the whole thing, it looks better IRL)

src/atoms/Button/buttons.module.scss
```diff
@@ -30,36 +30,25 @@ $brand-white: color('primary-2');
   width: 100%;
   outline: none;
   white-space: normal;
-  border-radius: 0;
```
kengorab, line 33: This was a duplicate style, `border-radius: 0;` is already applied on line 28

src/atoms/Button/buttons.module.scss
```diff
@@ -30,36 +30,25 @@ $brand-white: color('primary-2');
...
   &:focus {
     outline: none;
   }
 
-  .icon {
-    path {
-      transition: fill 300ms ease-out;
-      fill: $brand-white;
-    }
+  .icon path {
```
kengorab, line 39: It seems like someone had tried to style the button icons at one point, but this approach isn't exactly enough anymore. Since the `<Icon>` in a `<Button>` now renders the svg to the DOM, this styling works!

src/atoms/Button/buttons.module.scss
```diff
@@ -30,36 +30,25 @@ $brand-white: color('primary-2');
...
   }
 
   &.outline {
     background-color: transparent;
     color: $action-color;
     border: 2px solid inherit;
-  }
-}
 
-.disabled {
```
kengorab, line 54: I moved the `.disabled` styling down to the bottom, so it'd apply to any variant. Previously, an outlined + `variant="info"` button wouldn't appear disabled

src/atoms/Button/index.js
```diff
@@ -40,15 +40,15 @@ function Button( props ) {
         href={href}
         {...linkAttrs}
       >
-        { icon && <Icon icon={icon} className={styles['icon']} /> }
+        { icon && <Icon icon={icon} renderSVGDOM className={styles['icon']} /> }
```
kengorab, line 43: I'm curious to hear people's opinion on this. With the `renderSVGDOM` prop set here, the svg will be rendered directly into the DOM, as opposed to the other approach which used an `` tag with a `src` of the svg. There's no negative network implications of this (since the svg would have to be downloaded either way anyway), but I'm curious if there are any reasons why this approach was initially avoided. There could have maybe been performance implications with SVGs in older versions of React (esp large ones), but I don't think it should matter too much.

Sadly, there's no real scalable way to apply color to SVGs when they're brought in via an `` tag. There's the (well-supported) CSS `filter` property which can be used to apply coloring, but the API isn't exactly what we want. You can't just specify a hex value, you need to figure out how to represent your color in terms of CSS blend properties, and even that only works if you can assume that the SVG will always be loaded in as black (which isn't the case for all of our icons).

jmcolella: I thin this _should_ be fine, but it may cause issues with testing because the `renderSVGDOM` prop causes [a network request](https://github.com/policygenius/athenaeum/blob/master/src/atoms/Icon/index.js#L28), which I've had to mock out in the past.

I would say that's probably an OK trade-off since icons in buttons are not used _super_ frequently. Based on your comment, I would prefer to render the SVG to the DOM so we have maximum ability to manipulate it

src/atoms/Button/buttons.module.scss
jmcolella: bless 🙏

src/atoms/Button/index.js
jmcolella: also it probably wasn't baked in before because the `renderSVGDOM` prop was implemented a while after the `icon` prop for the `Button`. so it probably was just not backfilled

PR #831, opened 2020-04-08
Add discountGreenCircle-1 icon

mttrms:
Add the following icon to Athenaeum:

discountGreenCircle-1

### DIFF
--- a/src/atoms/Icon/constants.js
+++ b/src/atoms/Icon/constants.js
@@ -70,6 +70,7 @@ module.exports = {
     'discountColor',
     'discountGreen',
     'discountGreenCircle',
+    'discountGreenCircle-1',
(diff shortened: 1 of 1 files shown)


mttrms:
When running `gzip -9 -S 'z' newIcon.svg` and trying to 'view' your SVG you can't actually see it. I assumed you might need a server to decompress it or something, but couldn't find any info on that online or in the docs.

image


Reached out to Ken who was the last one to upload a new icon, and he pointed me to the separate docs in the wiki for adding new icons (Which Anders added after having similar issues following the docs)

If the error is expected I think it's worth updating the docs anyway just to confirm that this method creates an svgz which isn't viewable client side but 'works' when uploaded to the bucket.

PR #832, opened 2020-04-08
Update README with icon uploading docs

mttrms:
Previous README documentation had users using the following command: `gzip -9 -S 'z' newIcon.svg` which resulted in errors when actually viewing the SVGs locally.

[Separate docs](https://github.com/policygenius/athenaeum/wiki/Uploading-new-icons) are being used to upload icons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,12 +157,7 @@ Our RCL is structured around the concepts put forth by [Atomic Design](http://br
 
 ## Adding New Icons
 
-Upload the icons to our Google Cloud Bucket:
-
-1. gzip the svg
-  - `gzip -9 -S 'z' newIcon.svg`
-2. Upload `newIcon.svgz`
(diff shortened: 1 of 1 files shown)


gummoe:
What sort of errors were happening - do you know? I've never encountered any problems with the gzip command, but I might be missing something here.

mttrms:
When running `gzip -9 -S 'z' newIcon.svg` and trying to 'view' your SVG you can't actually see it. I assumed you might need a server to decompress it or something, but couldn't find any info on that online or in the docs.

image


Reached out to Ken who was the last one to upload a new icon, and he pointed me to the separate docs in the wiki for adding new icons (Which Anders added after having similar issues following the docs)

If the error is expected I think it's worth updating the docs anyway just to confirm that this method creates an svgz which isn't viewable client side but 'works' when uploaded to the bucket.

gummoe:
OH wow - who knew? Thanks for updating this!

PR #835, opened 2020-04-30
Fix: Update RadioField checked status and FullDateField proptypes

parky22:
1. Update `RadioField` component to use `input.checked` value if available. Otherwise, use the the current way of determining the checked status.
     - This is related to updating `react-final-form` to the latest version in `winnebago`: https://github.com/policygenius/winnebago/pull/578
2. Update the `FullDateField` component's `PropTypes` for `tooltip`. It was expecting on a `string`, but we oftentimes send functions that render a `ToolTip` component. 
     - This clears a warning in the console in `winnebago` at the very least

### DIFF
--- a/src/molecules/formfields/RadioField/index.js
+++ b/src/molecules/formfields/RadioField/index.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import camelCase from 'lodash/camelCase';
 import omit from 'lodash/omit';
+import isUndefined from 'lodash/isUndefined';
 import classnames from 'classnames';
 
 import styles from './radio_field.module.scss';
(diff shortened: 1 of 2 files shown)


src/molecules/formfields/RadioField/index.js
```diff
@@ -19,10 +20,13 @@ function RadioField(props) {
     className,
   ];
 
+  const isChecked = !isUndefined(input.checked) ? input.checked
```
jmcolella, line 23: can you explain a bit why we need to update this logic? didn't know there was a bug here but not using the `input.checked` attribute.

magcurt8: +1 I am not sure what the exact difference is here except for checking against undefined as well as checking for if the `input.value` matches the `radioValue`

magcurt8: +1 confused as well. It looks like the original code never checked for if `input.value` was defined before it checked if the `input.value` matched the `radioValue`, so I could see that leading to some issues, but not sure what `input.checked` vs. `input.value` does. Couldn't we just use `input.checked` and `!input.checked` in the `labelClasses` for if they are selected or not?

parky22: As discussed, because of how `react-final-form` components communicates with `athenaeum`, sometimes the `checked` attribute is not available. We will keep this as backwards compatible as possible but push towards using the `checked` value if possible.

PR #842, opened 2020-05-19
RF 61/add logged in indicator pc

bradwheel:
Add logged in indicator to components

### DIFF
--- a/src/molecules/AccountIndicator/account_indicator.module.scss
+++ b/src/molecules/AccountIndicator/account_indicator.module.scss
@@ -1,0 +1,119 @@
+$default-spacing: rem-calc(16px);
+
+.container {
+  position: relative;
+  max-width: rem-calc(500px);
+}
+
+.indicator {
+  width: auto;
+  display: flex;
+  justify-content: end;
+  align-items: center;
+  cursor: pointer;
+}
+
+.mobile-only {
+  display: none;
+}
+
+.desktop-only {
+  display: flex;
+}
+
+.close-icon {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: rem-calc(24px);
+  width: rem-calc(24px);
+  margin: $default-spacing;
+}
+
+.circle-initial {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  background: color('secondary-2');
+  border-radius: 100%;
+  height: rem-calc(32px);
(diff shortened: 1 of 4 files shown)


src/molecules/AccountIndicator/account_indicator.module.scss
```diff
@@ -0,0 +1,62 @@
...
+  height: rem-calc(32px);
+  width: rem-calc(32px);
+  margin: $default-spacing;
+}
+
+.dropdown {
+  position: absolute;
+  width: auto;
+  min-width: rem-calc(160px);
+  max-width: 300px;
```
jl08, line 33: should this also have `rem-calc` ?

src/molecules/AccountIndicator/index.js
```diff
@@ -0,0 +1,109 @@
...
+    const { open } = this.state;
+
+    return (
+      <div className={classnames(styles['container'])}>
+        <div onClick={this.handleIndicatorClick} className={classnames(styles['indicator'])}>
+          <div className={classnames(styles['circle-initial'])}>
+            <Text size={8} font='a' color='primary-2'>{this.getInitial(customerName)}</Text>
+          </div>
+
+          <Text size={9} font='a'>Hi, {this.getFirstName(customerName)}</Text>
```
jl08, line 56: should this display a different message (or not display at all) incase `getFirstName` returns empty string? or else it would say "Hi,  " 🤔

jl08 (commented):
looks great so far! just a few small questions

src/molecules/AccountIndicator/account_indicator.module.scss
bradwheel: probably? ill add one!

src/molecules/AccountIndicator/index.js
bradwheel: Ya definitely! Good catch. This should only display the text when they have a first name, otherwise just the icon. Changing!

src/molecules/AccountIndicator/account_indicator.module.scss
jl08: 👍

src/molecules/AccountIndicator/index.js
jl08: 👍

jl08 (approved):
🔥

PR #844, opened 2020-05-27
Add default props to AccountIndicator for links

bradwheel:
This commit adds default paths to logout and dashboard links in the
AccountIndicator component.

### DIFF
--- a/src/molecules/AccountIndicator/index.js
+++ b/src/molecules/AccountIndicator/index.js
@@ -209,4 +209,9 @@ AccountIndicator.propTypes = {
   onMenuLinkClick: PropTypes.func
 };
 
+AccountIndicator.defaultProps = {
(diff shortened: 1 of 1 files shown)


src/molecules/AccountIndicator/index.js
```diff
@@ -209,4 +209,9 @@ AccountIndicator.propTypes = {
   onMenuLinkClick: PropTypes.func
 };
 
+AccountIndicator.defaultProps = {
+  dashboardPath: '/users/dashboard',
```
trevornelson, line 213: I'm sure you're aware, but we'll just need to make sure to override this default for renter's.

bradwheel: good ol renters! :+1:

PR #850, opened 2020-08-05
add optional hideChat prop

whitneychoo:
[LIFE-826](https://policygenius.atlassian.net/browse/LIFE-826)

## Purpose
We want to hide the chat button on mobile, to give a better experience for users coming from the `Merlin` app, but also because mobile users don't really chat anyways

## Changes
- Added optional `hideChat` prop to the `Footer` component to conditionally show chat button on mobile
- Defaulted `hideChat` to false to not mess with other flows

### DIFF
--- a/src/templates/Footer/index.js
+++ b/src/templates/Footer/index.js
@@ -100,6 +100,7 @@ function renderPhoneInfo(phoneNumber, hours) {
 function Footer(props) {
   const {
     className,
+    hideChat,
     phoneNumber,
     emailAddress,
     links,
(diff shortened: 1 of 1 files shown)


src/templates/Footer/index.js
```diff
@@ -236,6 +239,12 @@ Footer.propTypes = {
    */
   className: PropTypes.string,
 
+  /**
+   * This prop will conditionally render the live chat button on mobile
+   */
```
mttrms, line 244: Should there be whitespace between the comment and the prop? Not super familiar with the syntax - but the other comments don't have that

whitneychoo: oh yeah i can update this!

PR #862, opened 2020-12-09
update datefield to support functions for tooltip

whitneychoo:
[LIFE-1065](https://policygenius.atlassian.net/browse/LIFE-1065)

Tooltips are not currently tracked in lifeweb analytics. This PR allows us to pass tooltips from life-web with all the necessary analytics events to athenaeum for rendering in form fields.

### DIFF
--- a/src/molecules/formfields/DateField/index.js
+++ b/src/molecules/formfields/DateField/index.js
@@ -52,9 +52,13 @@ DateField.propTypes = {
   fieldRef: PropTypes.func,
 
   /**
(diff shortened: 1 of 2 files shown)


jsonunger (approved):
Just curious, are there linting errors around PropTypes that forced the need for this change, or are we just being safe and adding it to account for the use case?

whitneychoo:
> Just curious, are there linting errors around PropTypes that forced the need for this change, or are we just being safe and adding it to account for the use case?

We are just adding it to support the tooltips analytics ticket!

PR #863, opened 2020-12-09
allow tooltip to be a func

whitneychoo:
[LIFE-1065](https://policygenius.atlassian.net/browse/LIFE-1065)

Tooltips are not currently tracked in lifeweb analytics. This PR allows us to pass tooltips from life-web with all the necessary analytics events to athenaeum for rendering in form fields.

### DIFF
--- a/src/molecules/formfields/DateField/index.js
+++ b/src/molecules/formfields/DateField/index.js
@@ -52,9 +52,13 @@ DateField.propTypes = {
   fieldRef: PropTypes.func,
 
   /**
-   * Adds a tooltip to the label. Provide string for text to be placed inside tooltip popup
+   * either a handler for clicking the tooltip, or text to go in the tooltip for the label
    */
-  tooltip: PropTypes.string,
+  tooltip: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.string,
+    PropTypes.object
+  ]),
 };
 
 export default DateField;
--- a/src/molecules/formfields/ShortDateField/index.js
+++ b/src/molecules/formfields/ShortDateField/index.js
@@ -73,9 +73,13 @@ ShortDateField.propTypes = {
   fieldRef: PropTypes.func,
 
   /**
-   * Adds a tooltip to the label. Provide string for text to be placed inside tooltip popup
+   * either a handler for clicking the tooltip, or text to go in the tooltip for the label
    */
-  tooltip: PropTypes.string,
+  tooltip: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.string,
+    PropTypes.object
+  ]),
 
   /**
    * Passes dayInputProps, monthInputProps, yearInputProps and fieldClass to provided function
--- a/src/molecules/formfields/util/BaseDateField/index.js
+++ b/src/molecules/formfields/util/BaseDateField/index.js
@@ -397,9 +397,13 @@ BaseDateField.propTypes = {
   fieldRef: PropTypes.func,
 
   /**
-   * Adds a tooltip to the label. Provide string for text to be placed inside tooltip popup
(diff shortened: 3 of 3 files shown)


jsonunger (approved):

jsonunger:
But PropTypes don't have any influence on how the logic works, it's just a way to help define what _should_ be passed in, but it doesn't prevent anything from being passed in, right? So with or without this code, the logic should still work. I'm in agreement that we should do this, just curious if there was anything blocking as a result of this code not being here.

whitneychoo:
> But PropTypes don't have any influence on how the logic works, it's just a way to help define what _should_ be passed in, but it doesn't prevent anything from being passed in, right? So with or without this code, the logic should still work. I'm in agreement that we should do this, just curious if there was anything blocking as a result of this code not being here.

So the Datefield component already supports rendering a tooltip given a function, but given the way we set the props up and their types, it only supported string. Previously I was getting an error that because the prop didn't match the type specified, it was not rendering and creating an error on the page. Without this, it actually wasn't working.

jsonunger:
Huh, that is surprising. I was under the impression PropTypes were just for documentation and will warn in development if the props don't match. I've never heard of prop types causing errors in rendering. Alright!

src/molecules/formfields/DateField/index.js
```diff
@@ -52,9 +52,13 @@ DateField.propTypes = {
...
 
   /**
-   * Adds a tooltip to the label. Provide string for text to be placed inside tooltip popup
+   * either a handler for clicking the tooltip, or text to go in the tooltip for the label
    */
-  tooltip: PropTypes.string,
+  tooltip: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.string,
+    PropTypes.object
```
showell215, line 60: Could you explain when the `tooltip` would be an object?

showell215 (commented):
I see there are a lot of field molecules that accept tooltips, and some of them ask for a `string` prop, and some have been updated to take `string`, `function`, or `object`. Should we take this opportunity to update all of them in case we run into this issue with text fields, or another field type? We could store the common propType definition in a util (maybe in that `fieldUtils` directory), and import it instead of redefining the propTypes in each consumer component. For example:

```
// util file
export const toolTipPropType = PropTypes.oneOf([PropTypes.func, PropTypes.string]);
```

Let me know what you think!

src/molecules/formfields/DateField/index.js
whitneychoo: yeah in the life-web PR you can see some examples but some tooltips were passed something that looks like this:
`{ children: "Don't include income from a spouse or partner", revealOnClick: true }`
in case other teams use tooltips in that same way, I just want to make sure we preserve this for them

showell215: oh, I see that now. Thanks!

whitneychoo:
> I see there are a lot of field molecules that accept tooltips, and some of them ask for a `string` prop, and some have been updated to take `string`, `function`, or `object`. Should we take this opportunity to update all of them in case we run into this issue with text fields, or another field type? We could store the common propType definition in a util (maybe in that `fieldUtils` directory), and import it instead of redefining the propTypes in each consumer component. For example:
> 
> ```
> // util file
> export const toolTipPropType = PropTypes.oneOf([PropTypes.func, PropTypes.string, PropTypes.func]);
> ```
> 
> Let me know what you think!

I'm hesitant to update it across the board for all tooltip props, but I'm happy to do it for other form fields things that I may have missed. Most of the time in life-web we use RCL components that don't have tooltips. If we need tooltips we tend to use the life-web version. The other reason why I'm hesitant to do it is that there has been talks about creating an upgraded RCL w/ design and tech overhaul so don't want to make too many 'just in case' changes. While it is not a big change, want to keep this work in the scope of the form field analytics ticket

PR #865, opened 2020-12-30
Add new icons

jsonunger:
Noting here, the `plant` and `piggyBank` icons look wrong, so I will circle back with Scott/design on Monday to get those fixed, but figured they could be in the repo and I can just update the SVG in the GCS bucket.

- additionalPolicy: 
- baby: 
- hands: 
- lunchPail: 
- piggyBank: 
- plant: 
- replacePolicy:


src/atoms/Icon/constants.js
```diff
@@ -163,6 +165,7 @@ module.exports = {
     'lock',
     'lockBlue',
     'lockOrange',
+    'lunchPail',
```
mttrms, line 168: LOL I think this is a briefcase but I like lunchpail

mttrms (approved):
LGTM!

src/atoms/Icon/constants.js
showell215: we could call it `rectangleWithHandle` to keep it generic? 😂

jsonunger: Screen Shot 2020-12-30 at 11 26 12 AM
I just used the name of the image in Figma

PR #869, opened 2021-02-03
Remove justify-content: end from AccountIndicator

danielnovograd:
The usage of `justify-content: end` has 'mixed support' according to a couple of big ol' warnings in the chrome console. Since we're also using `display: flex`, I don't think `end` does anything (since it is for non-flex children).

https://developer.mozilla.org/en-us/docs/Web/CSS/justify-content#syntax

### DIFF
--- a/src/molecules/AccountIndicator/account_indicator.module.scss
+++ b/src/molecules/AccountIndicator/account_indicator.module.scss
@@ -8,7 +8,6 @@ $default-spacing: rem-calc(16px);
 .indicator {
   width: auto;
   display: flex;
-  justify-content: end;
   align-items: center;
   cursor: pointer;
 }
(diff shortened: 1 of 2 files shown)


src/molecules/AccountIndicator/account_indicator.module.scss
```diff
@@ -8,7 +8,6 @@ $default-spacing: rem-calc(16px);
 .indicator {
   width: auto;
   display: flex;
-  justify-content: end;
```
maxnovak, line 11: Just double checking this is right, because this field is not deprecated according to the spec: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content

maxnovak: Agree that the behavior is wrong, just double checking rational.

danielnovograd: Not deprecated but "mixed support", which is enough for Chrome to show warnings about it. In this case, it's not applicable to the elements it's styling, so I think we need @bradwheel to weigh in on whether we need this to be `flex-end` or let it fall back to the default.

patrickskim (approved):
Lets merge it.

PR #871, opened 2021-02-10
Update node-sass version

maxnovak:
## Description
Upgrades node-sass version to get over security error and to allow it to run on node 14 w/o error.


danielnovograd (approved):
tis a beaut

jmcolella (approved):
@maxnovak LGTM

PR #873, opened 2021-02-23
MOR-1, MOR-2, MOR-3, MOR-4, MOR-5: Dummy v9 commit, in order to programmatically deploy v10

kengorab:
- In order to update to v10 in buildkite, I need to change the version
in the package.json to be `9.0.0`. This way, when the major version is
programmatically incremented, it'll be set to `10.0.0`.
- Note that this version will _NOT_ be deployed, as there is already a
deployment of version `9.0.0` in the wild

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "athenaeum",
-  "version": "8.26.1",
+  "version": "9.0.0",
   "description": "PolicyGenius React Component Library (RCL)",
   "main": "lib/index.js",
   "scripts": {


maxnovak (approved):
L̴͉̭͇̻̙͓͛̀G̴̛͙̰̻̊͊̓͛T̴͓̖̝̣̠̜͔͚̈́͜M̶̙͈̠̈́͆̅̓̓̃̊͜

patrickskim (commented):
The future. is now.

PR #874, opened 2021-02-25
MOR-9: Typography adjustments

kengorab:
- Update the `A` and `B` fonts in our typography. Some
of the styles are consolidated into one (eg. A6 and A7). In addition to
this consolidation, the `B` font will now be the serif font `Lora`.
- In addition to the style updates, I also updated the styleguidist
documentation a bit for the Branding section; now, the Typography section
just points right to the documentation for the Text component, and the
Spacer section points to the Spacer component.

### DIFF
--- a/src/assets/stylesheets/base_styles/_typography.scss
+++ b/src/assets/stylesheets/base_styles/_typography.scss
@@ -1,9 +1,9 @@
 @import 'colors';
 
-$font-family-serif:           'PolicyGenius', Georgia, serif;
-$font-family-sans-serif:      'PolicyGenius', Arial, sans-serif;
-$font-family-sans-serif-bold: 'PolicyGenius', Arial, sans-serif;
-$font-family-c: Georgia;
+@import url('https://fonts.googleapis.com/css2?family=Lora:wght@400;700&display=swap');
+
+$font-family-serif: Lora, Georgia, serif;
+$font-family-sans-serif: 'PolicyGenius', Arial, sans-serif;
 
 $main-font-color:             color('neutral-2');
 $label-font-color:            color('neutral-4');
@@ -29,7 +29,7 @@ $boldweight:                  900;
 }
 
 @mixin font-family-c() {
-  font-family: $font-family-c;
+  font-family: $font-family-serif;
   margin: 0;
 }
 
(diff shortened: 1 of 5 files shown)


src/assets/stylesheets/base_styles/_typography.scss
```diff
@@ -1,9 +1,9 @@
 @import 'colors';
 
-$font-family-serif:           'PolicyGenius', Georgia, serif;
-$font-family-sans-serif:      'PolicyGenius', Arial, sans-serif;
-$font-family-sans-serif-bold: 'PolicyGenius', Arial, sans-serif;
-$font-family-c: Georgia;
+@import url('https://fonts.googleapis.com/css2?family=Lora:wght@400;700&display=swap');
```
kengorab, line 3: The `'PolicyGenius'` font family is fetched from [this link](https://static.policygenius.com/fonts/PolicyGenius/font.css), and is just an alias for the `SharpSans` font. I didn't see a need for this layer of indirection for the new Lora font, especially since it is freely available and not a paid font.

maxnovak: Shouldn't we keep the font in that bucket just in case google decides to mess with the font file without telling anyone though?

patrickskim: @maxnovak I don't think we're deleting the older font. We wrapped the SharpSans font in Policygenius font, which added more complexity for not a lot of gain. 

```
Policygenius ->
 font weight: 400  -> SharpSans
 font weight: 900 -> SharpSans Display No 1 
```

patrickskim (approved):
🚀  🌔

src/assets/stylesheets/base_styles/_typography.scss
kengorab: As for your point though @maxnovak, it's valid though I don't think Google will just take down fonts once they've been made accessible. We could inter it into our own CDN to be extra extra safe, but I don't really think that's super necessary. And even if it did fail to load, we have a fallback (and even then, this font doesn't look _that_ different from the default Times New Roman anyway imo 😂 )

jmcolella (approved):
@kengorab LGTM

sunnymis (approved):
Lora Good To Me!

src/assets/stylesheets/base_styles/_typography.scss
patrickskim: Ack, I misunderstood. Lora in our bucket, 🤦  That said, Agree with Ken.

maxnovak (approved):
𝕷𝕲𝕿𝕸

PR #884, opened 2021-04-02
Allow empty alt text for icons

jsonunger:
[LIFE-1342](https://policygenius.atlassian.net/browse/LIFE-1342)

There are circumstances where it is appropriate and [recommended](https://www.w3.org/WAI/tutorials/images/decision-tree/) to provide an empty string for the `alt` attribute on an image tag. Right now, passing an empty string would result in the default behavior of using the icon name.

Instead, the change here allows an empty string to be passed and not overridden. Not providing an `alt` prop or passing `undefined` or `null` will keep the existing default functionality of using the icon name.

### DIFF
--- a/src/atoms/Icon/index.js
+++ b/src/atoms/Icon/index.js
@@ -102,7 +102,7 @@ class Icon extends React.Component {
             styles.img,
             isLazy && 'lazyload'
           )}
-          alt={alt || sanitizedIcon}
+          alt={alt != null ? alt : sanitizedIcon}
           title={title || sanitizedIcon}
           data-src={isLazy ? src : undefined}
           src={isLazy ? undefined : src}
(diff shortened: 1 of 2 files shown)


src/atoms/Icon/index.js
```diff
@@ -102,7 +102,7 @@ class Icon extends React.Component {
             styles.img,
             isLazy && 'lazyload'
           )}
-          alt={alt || sanitizedIcon}
+          alt={alt != null ? alt : sanitizedIcon}
```
showell215, line 105: it took me a good 20 seconds to remember that `null == undefined`

showell215 (approved):
👍

gummoe (approved):
LGTM!

PR #885, opened 2021-04-12
Soft Fail Snyk Step

nilaymalpekar:
Currently Snyk is used only to scan the vulnerabilities in repositories and it is not intended to block the pipelines. We noticed that if there are issues with Snyk API, the scan step fails and the pipeline gets blocked preventing out apps from deploying. To avoid such situations, this PR adds a soft fail to snyk step so that it does not block the pipeline if Snyk scan fails.


miles- (approved):
LGTM :rocket:

randy-concepcion (approved):

PR #894, opened 2021-05-17
new icons

showell215:
shield
Screen Shot 2021-05-17 at 2 16 37 PM

questionMarkBlackOutline
Screen Shot 2021-05-17 at 2 16 40 PM

### DIFF
--- a/src/atoms/Icon/constants.js
+++ b/src/atoms/Icon/constants.js
@@ -214,6 +214,7 @@ module.exports = {
     'priceTag',
     'questionBlackGreyCircle',
     'questionMark',
+    'questionMarkBlackOutline'
     'questionMarkWhite',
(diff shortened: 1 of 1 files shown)


gummoe (approved):

mttrms:
@showell215 i'm late to this since it was already merged in but I wonder if we should replace the pngs in the bucket with trimmed icons instead. I don't know if we'd re-use these but they're a little uneven.

showell215:
@mttrms are you referring to the screenshots? That's just from my sloppy screenshotting from the life growth page–  the assets are SVG

https://storage.googleapis.com/pg-static-assets/svgs/shield.svg
https://storage.googleapis.com/pg-static-assets/svgs/questionMarkBlackOutline.svg

PR #897, opened 2021-06-29
Added Icon

Scott-PG:
LIFE-1611
Added a fountainPen Icon for a life-web flow. (Preview image is a slightly off-center screengrab, just to make sure we can see it on a background, since the icon is all black)
Screen Shot 2021-06-29 at 3 47 32 PM


mttrms (approved):
🖋🚀

PR #898, opened 2021-07-20
Add value to checkbox htmlFor and id

jmcolella:
**Description**

* Add the `value` to the `htmlFor`/`id` props when rendering the `CheckBoxField`

**Context**
In order to make checkboxes unique when supplied a `value` in the input
prop, the component should use that variable when building the
htmlFor/id pairing for the label and input.

This commit adds the `value` prop to the htmlFor/id string if available.


parky22 (approved):
lgtm!

PR #905, opened 2021-12-06
sonarcloud

joseramosPG:
This PR migrates from SonarQube to SonarCloud. The logic of how and when it runs is the same. This simply points it to a new image.


jmcolella (approved):
@joseramosPG LGTM

showell215 (approved):

PR #913, opened 2022-02-14
adding security gating for the pipeline

M1Gd0nPG:
adding security gating to the pipeline to prevent critical and high vulns from being pushed to production


showell215 (approved):
👍

